PR #1, opened 2025-03-08
chore: added dx12 and vulkan as project dependencies via CMake

Narvin-Chana:
feat: added some basic placeholder code to work around and test the deps


Nowp:
This is a terrible pull request and I don't understand how in the world we could merge this...

PR #3, opened 2025-03-26
Backend creation

Narvin-Chana:
Simple graphics backend creation, includes passing in a windows handle (HWND), including example code for getting it from GLFW. Also changed CMakeLists.txt to define both API macros to 1 if valid and 0 if not valid (instead of just not defining them).


CMakeLists.txt
```diff
@@ -81,12 +91,16 @@ target_include_directories(Vex PUBLIC
 )
 
 # DX12 project files
-set(VEX_DX12_SOURCES
-)
+if (WIN32 AND VEX_ENABLE_DX12)
```
Narvin-Chana, line 94: In this file the main changes are that I've made the VEX_DX12 and VEX_VULKAN macros always be defined, they will instead be set to 0 or 1 according to the user's environment.

alexandreLBarrett (approved):

PR #4, opened 2025-03-29
Improved Logging

Narvin-Chana:
- Can now log to a file (will be renamed on program close)
- Added timestamps to logs
- Added log filter
- Fixed issue with the classic <windows.h> include

### DIFF
--- a/src/Vex/Logger.cpp
+++ b/src/Vex/Logger.cpp
@@ -1,0 +1,134 @@
+#include "Logger.h"
+
+#include <algorithm>
+#include <ctime>
+
+namespace vex
+{
+
+Logger::Logger()
+{
+    // Ensures the log file's path exists.
+    SetLogFilePath(filePath);
+}
+
+Logger::~Logger()
+{
+    CommitTimestampedLogFile();
+}
+
+void Logger::SetLogLevelFilter(LogLevel newFilter)
+{
+    GLogger.levelFilter = newFilter;
+}
+
+void Logger::SetLogFilePath(const std::filesystem::path& newLogFilePath)
+{
+    bool validPath = true;
+    std::filesystem::path strippedNewLogFilePath = std::filesystem::path(newLogFilePath).remove_filename();
+    if (!std::filesystem::exists(strippedNewLogFilePath))
+    {
+        validPath |= std::filesystem::create_directories(strippedNewLogFilePath);
+    }
+
+#if defined(_WIN32)
+    static constexpr size_t MaxPathLength = 256;
+    // Windows only allows for paths shorter than 256.
+    // If this is the case, we give up on changing the log file name.
+    if (strippedNewLogFilePath.string().size() >= MaxPathLength)
+    {
+        VEX_LOG(Error, "Unable to set log path because the total path length would be greater than 256.");
+        return;
+    }
+#endif
+
+    if (validPath)
+    {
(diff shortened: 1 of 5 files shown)


src/Vex/Logger.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+void Logger::CloseLogFile()
+{
+    if (!logOutput.has_value())
+    {
+        return;
+    }
+
+    logOutput->close();
```
alexandreLBarrett, line 84: no need to close RAII handles it

src/Vex/Logger.cpp
```diff
@@ -0,0 +1,133 @@
...
+    std::string fileName = std::string(LogFileName).append(std::format("_{}", timeStamp)).append(LogFileFormat);
+    std::filesystem::path timeStampedFilePath = std::filesystem::path(filePath).remove_filename() / fileName;
+
+#if defined(_WIN32)
+    // Windows only allows for paths shorter than 256.
+    // If this is the case, we give up on renaming the log file.
+    static constexpr size_t MaxPathLength = 256;
+    if (timeStampedFilePath.string().size() >= MaxPathLength)
+    {
+        return;
```
alexandreLBarrett, line 111: maybe add a log here and above for visibility

src/Vex/Logger.h
```diff
@@ -18,36 +27,103 @@ enum LogLevel : u8
...
+{
+    Logger();
+    ~Logger();
+
+    template <class... Args>
+    void Log(LogLevel level, std::format_string<Args...> formatMessage, Args&&... args)
+    {
+        std::string userMessage = std::format(formatMessage, std::forward<Args>(args)...);
+        std::string timeStampedFormatMessage =
+            std::format("[{}][{}] {}", GetTimestampString(), LogLevelToString(level), userMessage);
```
alexandreLBarrett, line 67: could move userMessage here since you dont use it later

src/Vex/Logger.h
```diff
@@ -18,36 +27,103 @@ enum LogLevel : u8
...
+
+        if (destinationFlags & Console)
+        {
+            std::println("{}", timeStampedFormatMessage);
```
alexandreLBarrett, line 71: We might want to favor cout here since it doesn't flush the buffer. We should check if println does it

alexandreLBarrett: I wrote this before seeing whats below. If you think this is critical it can stay as is

src/Vex/Logger.h
```diff
@@ -18,36 +27,103 @@ enum LogLevel : u8
...
+private:
+    void OpenLogFile();
+    void CloseLogFile();
+
+    // Closes the stream and renames the log file with the current timestamp.
+    void CommitTimestampedLogFile();
+
+    static constexpr const char* LogFileName = "vex";
+    static constexpr const char* LogFileFormat = ".log";
+    static constexpr const char* LogFileNameFormat = "vex.log";
```
alexandreLBarrett, line 104: use the constants above to compute this? Could use std::string_view as i'm pretty sure they can be constexpr

src/Vex/Logger.h
Narvin-Chana: Println does not flush (appends a '\n', without flushing the buffer)

alexandreLBarrett (approved):

PR #6, opened 2025-03-30
Unique Handle finished implementation

Narvin-Chana:


alexandreLBarrett:
Might be a bit late about this. We could've use unique_ptr with custom deleter no? instead of copy pasting impl of unique_ptr. It's fine if you think it could be limiting this way. I figure it's a bit more idiomatic and would save us from reinventing the wheel

PR #7, opened 2025-03-30
Feature checker

Narvin-Chana:
Changes how graphics backend is instantiated, moves api-specific logic outside of the Vex/ folder.

The Vex.h and Vex.cpp files allow the user to create a graphics backend by just passing in an enum (simpler usage than passing in a pointer on the client-side).


src/Vex/UniqueHandle.h
```diff
@@ -1,187 +1,27 @@
...
-    explicit operator bool() const noexcept
-    {
-        return ptr != nullptr;
-    }
-
-private:
-    T* ptr = nullptr;
-};
+template <class T, class Deleter = std::default_delete<T>>
+using UniqueHandle = std::unique_ptr<T, Deleter>;
```
alexandreLBarrett, line 11: You gave up on the naming convention then?

src/Vex/UniqueHandle.h
```diff
@@ -1,187 +1,27 @@
...
     requires(not std::is_array_v<T>)
 constexpr UniqueHandle<T> MakeUnique(Args&&... args)
 {
-    return UniqueHandle<T>(new T(std::forward<Args>(args)...));
+    return std::make_unique<T>(std::forward<Args>(args)...);
 }
 
 template <class T>
     requires(std::is_unbounded_array_v<T>)
 constexpr UniqueHandle<T> MakeUnique(std::size_t size)
```
alexandreLBarrett, line 22: I dont think constepxr here will do anything. You can't to my knowledge heap allocation in a constexpr context. It will only be called at runtime

alexandreLBarrett (approved):

src/Vex/UniqueHandle.h
Narvin-Chana: Yeee... If I wanted the correct naming, I'd have to write a wrapper around unique_ptr... and at that point is it really a worthwhile time commitment? Probably not haha

src/Vex/UniqueHandle.h
Narvin-Chana: True, will remove

PR #8, opened 2025-04-05
Physical device abstraction

Narvin-Chana:
**!!!IMPORTANT!!!
DEPENDS ON #7 
!!!IMPORTANT!!!**

- Adds actual physical device abstraction
- Can query the RHI to get the list of physical devices
- The backend then selects one (could add some user input here if needed later on), then initializes the RHI with it.
- The RHI is then free to create its logical device and swapchain (not implemented in this MR)


src/Vulkan/VkRHI.cpp
```diff
@@ -0,0 +1,174 @@
...
+        {
+            // Device local implies its VRAM memory
+            VkDeviceSize heapSize = memoryProperties.memoryHeaps[i].size;
+
+            // Convert to a more readable format (MB)
+            totalDeviceLocalMemoryMB += static_cast<double>(heapSize) / (1024.0 * 1024.0);
+        }
+    }
+    return totalDeviceLocalMemoryMB;
+}
```
alexandreLBarrett, line 66: shouldn't this be a virtual method on the PhysicalDevice?

Narvin-Chana: Good point! Moved these initializers on both the dx and vk side inside the platform specific physical devices.

alexandreLBarrett (approved):

PR #9, opened 2025-04-06
chk and chkSoft equivalent for vulkan

alexandreLBarrett:


Narvin-Chana (changes requested):
Looks good to me, just an issue with the name of the defines you added.

Narvin-Chana (approved):

PR #14, opened 2025-04-19
Window Resize and Fullscreen Toggle

Narvin-Chana:
- Moved boilerplate glfw code to an ExampleApplication that can be reused by examples to avoid duplication
- Added DX12RHI logging of "leaked" (in reality this prints out those that are still alive) resources at destruction

### DIFF
--- a/examples/ExampleApplication.cpp
+++ b/examples/ExampleApplication.cpp
@@ -14,7 +14,7 @@
 #include <Vex.h>
 #include <Vex/Logger.h>
 
-ExampleApplication::ExampleApplication()
+ExampleApplication::ExampleApplication(const std::string& windowName)
 {
 #if defined(__linux__)
     glfwInitHint(GLFW_PLATFORM, GLFW_PLATFORM_X11);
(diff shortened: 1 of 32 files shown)


examples/ExampleApplication.cpp
```diff
@@ -0,0 +1,105 @@
...
+    height = DefaultHeight;
+
+    windowedInfo = { .width = DefaultWidth, .height = DefaultHeight };
+    glfwGetWindowPos(window, &windowedInfo.x, &windowedInfo.y);
+
+    static std::function<void(GLFWwindow*, int, int)> ResizeCallback = [this](auto* w, auto width, auto height)
+    { this->OnResize(w, width, height); };
+
+    // Set static callback that uses the function object
+    glfwSetWindowSizeCallback(window, [](GLFWwindow* w, int width, int height) { ResizeCallback(w, width, height); });
```
alexandreLBarrett, line 43: Could use std::bind to save yourself a static lambda

alexandreLBarrett (approved):

examples/ExampleApplication.h
```diff
@@ -1,13 +1,15 @@
...
+
 #include <Vex.h>
 
 class GLFWwindow;
 
 class ExampleApplication
 {
 public:
-    ExampleApplication();
+    ExampleApplication(const std::string& windowName);
```
alexandreLBarrett, line 12: could use string_view if we copy it later. Saves a string creation

PR #15, opened 2025-04-23
Implemented swapchain on vulkan side. Resizing should also work.

alexandreLBarrett:


Narvin-Chana:
Ubuntu (gcc) errors to fix

src/Vulkan/VkMemory.h
```diff
@@ -0,0 +1,23 @@
+﻿#pragma once
+
+#include "Vex/Logger.h"
+#include "VkHeaders.h"
+
+namespace vex::vk
+{
+inline uint32_t getBestMemoryType(::vk::PhysicalDevice device, uint32_t typeFilter, ::vk::MemoryPropertyFlags flags)
```
Narvin-Chana, line 8: u32 (return type and second parameter) and GetBestMemoryType

src/Vulkan/VkGPUContext.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include "VkHeaders.h"
+
+namespace vex::vk
+{
+struct VkCommandQueue;
+}
+namespace vex::vk
+{
+
```
Narvin-Chana, line 11: (code style nitpick) could just forward declare inside the same namespace

src/DX12/DX12SwapChain.h
```diff
@@ -24,6 +24,7 @@ class DX12SwapChain : public RHISwapChain
                   const ComPtr<ID3D12CommandQueue>& commandQueue,
                   const PlatformWindow& platformWindow);
     virtual ~DX12SwapChain() override;
+    virtual void AcquireNextBackbuffer(u8 frameIndex);
```
Narvin-Chana, line 27:
```suggestion
    virtual void AcquireNextBackbuffer(u8 frameIndex) override;
```

src/Vulkan/VkTexture.h
```diff
@@ -0,0 +1,35 @@
...
+
+    ::vk::Image image;
+};
+
+class VkTexture final : public RHITexture
+{
+public:
+    // Takes ownership of the image
+    VkTexture(const TextureDescription& description, ::vk::UniqueImage rawImage);
+    VkTexture(TextureDescription&& description, ::vk::UniqueImage rawImage);
```
Narvin-Chana, line 23: I wonder if there could there be a better way to indicate that ownership has been taken? UniqueImage I'm guessing is just a handle, so std::move wouldn't really be appropriate.

Narvin-Chana (changes requested):
Nice stuff! I'd like to see the MR after the window resize PR to make sure resizing works as expected

src/Vulkan/VkTexture.h
alexandreLBarrett: UniqueHandle is basically a unique ptr. It can't be copied. So this forces users to move into it

Narvin-Chana: I was talking about ::vk::UniqueImage, not UniqueHandle (unless I'm misunderstanding your comment?).

Narvin-Chana (approved):

PR #16, opened 2025-04-27
Command context (this is a big one)

Narvin-Chana:
Contains the minimum stuff to draw a triangle using only compute shaders!

Lots of code here, the main features:
- Command context which is the user-facing API, the command context will register commands and then submit them to the API via an internal RHICommandList
- The starting skeleton for our user-facing textures (including a new FreeList container to map handles to the actual underlying RHI resource).
- A new RHIResourceLayout class that instantiates internal API resource mapping. This is akin to DX12's root signature/Vulkan's Pipeline Layouts/Descriptor Set Layouts
- A way to compile pipeline state objects, this should be way easier in Vulkan since dynamic rendering exists.
- Currently the shader I use was manually compiled outside of Vex and passed in directly as bytecode. This was to allow myself to implement the strict minimum of code to get a hello triangle working.

Lots of stuff is hardcoded in (eg: the texture we render to is determined inside the resource layout) but I felt this was becoming too large to not put to a review.


examples/CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ foreach(EXAMPLE_DIR ${EXAMPLE_DIRS})
 
     # Only create the example if there are source files.
     if (EXAMPLE_SOURCES)
-        add_executable(${EXAMPLE_NAME} ${EXAMPLE_SOURCES})
+        add_executable(${EXAMPLE_NAME} ${EXAMPLE_SOURCES} "ExampleApplication.h" "ExampleApplication.cpp")
```
alexandreLBarrett, line 34: Why not add them to EXAMPLE_SOURCES?

src/DX12/DX12CommandList.cpp
```diff
@@ -66,12 +70,171 @@ void DX12CommandList::Close()
...
+    ID3D12Resource* uavTexture = reinterpret_cast<const DX12ResourceLayout&>(layout).uavTexture.Get();
+
+    {
+        D3D12_RESOURCE_BARRIER uavToSource = CD3DX12_RESOURCE_BARRIER::Transition(uavTexture,
+                                                                                  D3D12_RESOURCE_STATE_UNORDERED_ACCESS,
+                                                                                  D3D12_RESOURCE_STATE_COPY_SOURCE);
+        D3D12_RESOURCE_BARRIER backbufferToDest =
+            CD3DX12_RESOURCE_BARRIER::Transition(bb, D3D12_RESOURCE_STATE_PRESENT, D3D12_RESOURCE_STATE_COPY_DEST);
+        D3D12_RESOURCE_BARRIER firstBarriers[] = { uavToSource, backbufferToDest };
+        commandList->ResourceBarrier(2, firstBarriers);
```
alexandreLBarrett, line 223: this should be easy to abstract a little more on the RHI level. Could be done at a later time tho

src/Vex/Handle.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include <Vex/Types.h>
+
+namespace vex
+{
+template <class Derived>
+struct Handle
+{
+    // First 24 bits indicate handle index (max of 16'777'215), following 8 bits indicate the generation (max of
```
alexandreLBarrett, line 10: What would you use generation for?

src/Vex/RHI/RHIPipelineState.h
```diff
@@ -0,0 +1,73 @@
...
+    u32 rootSignatureVersion = 0;
+    u32 computeShaderVersion = 0;
+};
+
+} // namespace vex
+
+// clang-format off
+VEX_MAKE_HASHABLE(vex::GraphicsPipelineStateKey, 
+    VEX_HASH_COMBINE(seed, obj.vertexShader);
+    VEX_HASH_COMBINE(seed, obj.pixelShader);
```
alexandreLBarrett, line 67: using ; here? Is this intentional. I see that the macro is variadic and this doesn't use that feature

examples/CMakeLists.txt
Narvin-Chana: Example sources is a GLOBBED result, not a manually defined constant, idk how to append to another variable in cmake, but in any case shouldn't change much.

src/Vex/Handle.h
Narvin-Chana: I've improved this further in the following PR, for now consider this to just be a baseline.

PR #17, opened 2025-05-11
Texture Abstraction w/ DX12 Impl

Narvin-Chana:
- Can now copy the contents of a texture to another (as long as they have the same format/size)
- Handling of bindless indices and resource views on the DX12 side, I've left these voluntarily very API-specific as the underlying logic could vary quite a bit for Vulkan.
- Registering of textures and mapping between RHITexture and Texture.
- A LOT of todos in code, but again this PR could have been 5x larger if I had to do everything.
- The app still crashes upon resize, but this is normal and related to us not handling resource lifetimes, when we add ResourceCleanup logic, we'll be able to resize without issues.


src/DX12/DX12CommandList.cpp
```diff
@@ -173,20 +165,19 @@ void DX12CommandList::SetLayoutLocalConstants(const RHIResourceLayout& layout,
         return;
     }
 
-    std::vector<u8> padding =
-        std::vector<u8>((globalRootSignature.GetMaxLocalConstantSize() - finalSize) * sizeof(DWORD));
+    // Padding to fill out the unused local constants space.
+    std::vector<u8> padding = std::vector<u8>((dxResourceLayout.GetMaxLocalConstantSize() - finalSize) * sizeof(DWORD));
```
alexandreLBarrett, line 169: could resize the dataToUpload to always match the max so that the padding is implied in the buffer. It might be a special case idk. If not it would make the code less complicated

src/DX12/DX12CommandList.cpp
```diff
@@ -197,7 +188,127 @@ void DX12CommandList::SetLayoutLocalConstants(const RHIResourceLayout& layout,
     }
 }
 
-void DX12CommandList::Dispatch(const std::array<u32, 3>& groupCount, RHIResourceLayout& layout, RHITexture& backbuffer)
+static TextureViewType GetTextureViewType(const ResourceBinding& binding)
```
alexandreLBarrett, line 191: we could use anonymous namespaces maybe. It's more ceeplusplusee but not a big deal. I personnally find it a bit cleaner

src/DX12/DX12CommandList.cpp
```diff
@@ -197,7 +188,127 @@ void DX12CommandList::SetLayoutLocalConstants(const RHIResourceLayout& layout,
...
+    // Now we can bind the bindless textures as constants in our root constants!
+    // TODO: figure out how this interacts with local root constants, there should be a way to get the first slot we can
+    // write bindless indices to (that is after local constants). For now we just default to slot 0, and suppose that no
+    // constants exist (just to get our Hello Triangle working).
+    switch (type)
+    {
+    case CommandQueueType::Graphics:
+        commandList->SetGraphicsRoot32BitConstants(0, bindlessHandles.size(), bindlessHandles.data(), 0);
+    case CommandQueueType::Compute:
+        commandList->SetComputeRoot32BitConstants(0, bindlessHandles.size(), bindlessHandles.data(), 0);
```
alexandreLBarrett, line 297: would this collide with anything set in SetLayoutConstants?

src/DX12/DX12Shader.h
```diff
@@ -495,32 +535,37 @@ inline const u8 TriangleComputeShader[] = {
...
-    0x1d, 0x8e, 0x20, 0xd2, 0x60, 0xf8, 0x2a, 0x10, 0x74, 0x96, 0x41, 0x08, 0x82, 0xe1, 0x08, 0x65, 0x0d, 0x8a, 0x6f,
-    0xb8, 0x41, 0x41, 0x83, 0xa1, 0x02, 0x41, 0x87, 0x23, 0x12, 0x37, 0x28, 0xbe, 0x0a, 0x04, 0x9d, 0x65, 0x10, 0x86,
-    0xc0, 0xac, 0x4b, 0x3e, 0x06, 0x07, 0x41, 0x7c, 0x46, 0x0c, 0x1c, 0x00, 0x04, 0xc1, 0xa0, 0x41, 0x85, 0x3d, 0x08,
-    0x83, 0x31, 0x10, 0x03, 0x3d, 0xc8, 0xb0, 0x20, 0x0e, 0xe8, 0x60, 0x96, 0x80, 0x18, 0x31, 0x70, 0x00, 0x10, 0x04,
-    0x83, 0x06, 0x15, 0xf6, 0x20, 0x0c, 0xc6, 0x40, 0x0c, 0xf4, 0x00, 0x0e, 0xe0, 0x00, 0x0e, 0xe2, 0x80, 0x0e, 0x66,
-    0x09, 0x08, 0x04, 0x00, 0x00, 0x00, 0x00, 0x00
+    0x40, 0x1c, 0xe4, 0x01, 0x00, 0x00, 0x00, 0x71, 0x20, 0x00, 0x00, 0x24, 0x00, 0x00, 0x00, 0x06, 0x60, 0x70, 0xac,
+    0x09, 0x20, 0x8d, 0x19, 0x50, 0xc3, 0xe5, 0x3b, 0x8f, 0x0f, 0x34, 0x8d, 0x33, 0x01, 0x13, 0x11, 0x02, 0xcd, 0xb0,
+    0x10, 0x56, 0xb0, 0x0d, 0x97, 0xef, 0x3c, 0xbe, 0x10, 0x50, 0x45, 0x41, 0x44, 0xa5, 0x03, 0x0c, 0x25, 0x61, 0x00,
+    0x02, 0xe6, 0x23, 0xb7, 0x6d, 0x07, 0xdd, 0x70, 0xf9, 0xce, 0xe3, 0x0b, 0x11, 0x01, 0x4c, 0x44, 0x08, 0x34, 0xc3,
```
alexandreLBarrett, line 541: I think these should be prime number and they aren't 👎

src/DX12/DX12Texture.h
```diff
@@ -1,26 +1,92 @@
...
+    bool operator==(const DX12TextureView&) const = default;
+};
+
+} // namespace vex::dx12
+
+// clang-format off
+VEX_MAKE_HASHABLE(vex::dx12::DX12TextureView,
+    VEX_HASH_COMBINE_ENUM(seed, obj.type);
+    VEX_HASH_COMBINE_ENUM(seed, obj.dimension);
+    VEX_HASH_COMBINE_ENUM(seed, obj.format);
```
alexandreLBarrett, line 38: nitpick: could detect that they are enums in the HASH_COMBINE variant. This would allow for every call to be the same

src/Vex/Handle.h
```diff
@@ -11,6 +11,12 @@ struct Handle
     // 255).
     u32 value = ~0U;
```
alexandreLBarrett, line 12: you could use bitfields here to save some code complexity

```
u32 value : 24;
u8 generation : 8;
```

Narvin-Chana: Generally bitfields are less portable so I try to avoid using them. There are weird situations where the example you gave can be 8 bytes large (source: Albin). I agree it would simplify the code, but the risk in portability doesn't seem worthwhile imo.

src/DX12/DX12CommandList.cpp
Narvin-Chana: I feel like between anonymous namespaces and static functions, there is no real difference. The typical reason to use private (named) namespaces is when you use Unity build, to avoid name conflicts. Functions in anonymous namespaces, even if declared separately, are considered as part of the same namespace (atleast according to tests I've done in Wandbox). 

IMO we should prefer named private namespaces with the following convention: namespace FILENAME_Internal or FILENAME_Private. Is this ok w/ you, and if so which do you prefer?

src/DX12/DX12CommandList.cpp
Narvin-Chana: Probably, layout local/global constants binding is not yet completely finalized, I'll add a VEX_NOT_IMPLEMENTED to them to indicate this.

src/DX12/DX12Texture.h
Narvin-Chana: How would you detect that they are an enum? Since this is all done macro-time?

alexandreLBarrett: use type trait in the macro body perhapse. I don't remember how they are made

src/DX12/DX12CommandList.cpp
Narvin-Chana: I have no real preference for using Unity build right now, I think it could possibly cause issues if users of Vex use Unity build though?

PR #18, opened 2025-05-18
Texture resource state and cleanup

Narvin-Chana:
- Adds resource states for textures
- Adds a way to cleanup resources after deletion, while still allowing them to be used by the GPU frames in flight.

### DIFF
--- a/src/DX12/DX12CommandList.cpp
+++ b/src/DX12/DX12CommandList.cpp
@@ -7,6 +7,7 @@
 #include <DX12/DX12Formats.h>
 #include <DX12/DX12PipelineState.h>
 #include <DX12/DX12ResourceLayout.h>
+#include <DX12/DX12States.h>
 #include <DX12/DX12Texture.h>
 #include <DX12/HRChecker.h>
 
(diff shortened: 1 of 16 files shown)


src/DX12/DX12CommandList.cpp
```diff
@@ -323,50 +354,15 @@ void DX12CommandList::Dispatch(const std::array<u32, 3>& groupCount)
...
 void DX12CommandList::Copy(RHITexture& src, RHITexture& dst)
 {
-    // TODO: handle resource states (I've created a trello for this)
-
+    VEX_ASSERT(src.GetDescription().width == dst.GetDescription().width &&
+                   src.GetDescription().height == dst.GetDescription().height &&
+                   src.GetDescription().depthOrArraySize == dst.GetDescription().depthOrArraySize &&
+                   src.GetDescription().mips == dst.GetDescription().mips &&
+                   src.GetDescription().format == dst.GetDescription().format,
+               "The two textures must be compatible in order to Copy to be useable.");
```
alexandreLBarrett, line 362: we could add a custom == operator on the RHITexture or some kind of function maybe

src/DX12/DX12States.cpp
```diff
@@ -0,0 +1,42 @@
...
+namespace vex::dx12
+{
+
+D3D12_RESOURCE_STATES RHITextureStateToDX12State(RHITextureState::Flags state)
+{
+    D3D12_RESOURCE_STATES outState = D3D12_RESOURCE_STATE_COMMON;
+    if (state & RHITextureState::RenderTarget)
+    {
+        outState |= D3D12_RESOURCE_STATE_RENDER_TARGET;
+    }
```
alexandreLBarrett, line 12: this feels like it could be a one liner. Too lazy to figure it out

src/Vex/Containers/ResourceCleanup.h
```diff
@@ -0,0 +1,26 @@
...
+#include <Vex/Types.h>
+#include <Vex/UniqueHandle.h>
+
+namespace vex
+{
+
+class ResourceCleanup
+{
+public:
+    using TCleanupVariant = std::variant<UniqueHandle<RHITexture>, UniqueHandle<RHIBuffer>>;
```
alexandreLBarrett, line 17: Unreal T prefix should be used everywhere or nowhere. I don't think it's necessary

src/DX12/DX12CommandList.cpp
Narvin-Chana: This was voluntarily not done using a == operator, these are the fields that D3D12 requires to be identical on two resources in order to allow copying from one to the other. Other fields like the usage or clear value are not required to be identical, this could obviously differ in Vulkan so I think its acceptable to keep this as-is.

src/DX12/DX12States.cpp
Narvin-Chana: Certainly, I think the code as-is probably is good enough, sometimes ease of reading is better than one liner magic :)

src/Vex/Containers/ResourceCleanup.h
Narvin-Chana: Should our convention then be to not include T at all (if so fine by me)? This might be the first example of an inner-class "using" in the project.

PR #22, opened 2025-07-09
Hello triangle now working in Vulkan!

alexandreLBarrett:


src/Vulkan/VkCommandList.cpp
```diff
@@ -79,59 +100,305 @@ void VkCommandList::SetPipelineState(const RHIGraphicsPipelineState& graphicsPip
...
+                                      0u,
+                                      [](u32 acc, const ConstantBinding& binding) { return acc + binding.size; });
+
+    VEX_ASSERT(total <= pushConstantData.size(),
+               "Unable to bind local constants, you have surpassed the limit Vulkan allows for in root signatures.");
+
+    u8 currentIndex = 0;
+    for (const auto& binding : constants)
+    {
+        std::uninitialized_copy_n(static_cast<u8*>(binding.data), binding.size, &pushConstantData[currentIndex]);
```
Narvin-Chana, line 129: What's the gain of using std::uninitialized_copy when in any case, your initialization of the vector will have initialized the chars to 0? I would just use a memcpy and call it a day.

src/Vulkan/VkCommandList.cpp
```diff
@@ -79,59 +100,305 @@ void VkCommandList::SetPipelineState(const RHIGraphicsPipelineState& graphicsPip
...
-    VEX_NOT_YET_IMPLEMENTED();
+    auto& vkResourceLayout = reinterpret_cast<const VkResourceLayout&>(layout);
+    auto& vkDescriptorPool = reinterpret_cast<VkDescriptorPool&>(descriptorPool);
+
+    std::vector<BindlessHandle> bindlessHandles;
+    bindlessHandles.reserve(textures.size() + buffers.size());
+
+    for (auto& [binding, usage, rhiTexture] : textures)
+    {
+        auto dxTexture = reinterpret_cast<VkTexture*>(rhiTexture);
```
Narvin-Chana, line 166: You probably want to name this vkTexture instead of dxTexture.

src/Vulkan/VkCommandList.cpp
```diff
@@ -79,59 +100,305 @@ void VkCommandList::SetPipelineState(const RHIGraphicsPipelineState& graphicsPip
...
+    }
+
+    ::vk::ShaderStageFlags stageFlags;
+    switch (type)
+    {
+    case CommandQueueTypes::Compute:
+        stageFlags = ::vk::ShaderStageFlagBits::eCompute;
+        break;
+    case CommandQueueTypes::Graphics:
+        stageFlags = ::vk::ShaderStageFlagBits::eAllGraphics;
```
Narvin-Chana, line 195: With this flag will a graphics command queue still be able to do compute? (hopefully)

src/Vulkan/VkCommandList.cpp
```diff
@@ -79,59 +100,305 @@ void VkCommandList::SetPipelineState(const RHIGraphicsPipelineState& graphicsPip
...
-    VEX_NOT_YET_IMPLEMENTED();
+    auto& vkTexture = reinterpret_cast<VkTexture&>(texture);
+    auto memBarrier = GetMemoryBarrierFrom(vkTexture, newState);
+
+    commandBuffer->pipelineBarrier2({ .imageMemoryBarrierCount = 1, .pImageMemoryBarriers = &memBarrier });
+
+    texture.SetCurrentState(newState);
 }
 
+// we only transfer access here and not layout
```
Narvin-Chana, line 310: What's the difference between transferring access and layout in Vk?

src/Vex/RHI/RHIDescriptorPool.h
```diff
@@ -2,14 +2,30 @@
...
 {
+    enum class Type : i8
+    {
+        None = -1,
+        UniformBuffer = 0,
+        StorageBuffer = 1,
+        Texture = 2,
+        StorageImage = 3,
+        Num = 4
+    } type = Type::None;
```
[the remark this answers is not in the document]
Narvin-Chana: To add to this, the concepts seem to not map well to DX12, where all types of descriptor handles are inside a singular heap. Also UniformBuffer/StorageBuffer/StorageImage I'm guessing match with: SRV Buffer / UAV Buffer / UAV Texture?

src/Vulkan/VkRHI.cpp
```diff
@@ -275,24 +288,44 @@ UniqueHandle<RHIGraphicsPipelineState> VkRHI::CreateGraphicsPipelineState(const
...
+    const auto& graphicsQueue = commandQueues[CommandQueueTypes::Compute];
+
+    ::vk::SubmitInfo submitInfo{ .commandBufferCount = 1, .pCommandBuffers = &*vkCmdList.commandBuffer };
+    VEX_VK_CHECK << graphicsQueue.queue.submit(1, &submitInfo, nullptr);
+
+    // This shouldn't be needed i think
+    // TODO: Find a way to make sure this isn't needed. It screams otherwise because the commandlist is supposedly in
+    // use
+    VEX_VK_CHECK << graphicsQueue.queue.waitIdle();
+
```
Narvin-Chana, line 323: Oh wait what? We probably don't want to open and submit an entire command list just for a texture's creation, is it not possible to specify an initial state in Vk? If not, I'd just leave it in the default state and only transition when needed (eg: for UAV state, only when the dispatch is called in command context).

src/Vulkan/VkRHI.cpp
```diff
@@ -314,7 +347,7 @@ void VkRHI::ExecuteCommandList(RHICommandList& commandList)
 
     ::vk::SemaphoreSubmitInfo semSignalInfo{
         .semaphore = *cmdQueue.waitSemaphore,
-        .value = cmdQueue.waitValue + 1,
+        .value = ++cmdQueue.waitValue,
```
Narvin-Chana, line 350: oof xD

src/Vulkan/VkRHI.cpp
```diff
@@ -363,10 +396,17 @@ void VkRHI::WaitFence(CommandQueueType queueType, RHIFence& fence, u32 fenceInde
 
     VEX_VK_CHECK << commandQueues[queueType].queue.submit(submit);
 }
+
+void VkRHI::ModifyShaderCompilerEnvironment(std::vector<const wchar_t*>& args, std::vector<ShaderDefine>& defines)
+{
+    args.push_back(L"-spirv");
+    defines.emplace_back(L"VEX_VULKAN", L"");
```
Narvin-Chana, line 403: For the defines, it'd be nice if you could add a constructor with an empty string default parameter for the define value.

src/Vulkan/VkTexture.cpp
```diff
@@ -76,3 +185,47 @@ void VkTexture::CreateImage(VkGPUContext& ctx)
...
+    {
+    case Common:
+        return ::vk::ImageLayout::eUndefined;
+        break;
+    case RenderTarget:
+        return ::vk::ImageLayout::eColorAttachmentOptimal;
+        break;
+    case UnorderedAccess:
+        return ::vk::ImageLayout::eGeneral;
+        break;
```
Narvin-Chana, line 206: General is the state for unordered access read/write in vk? If so that's wack.

src/Vulkan/VkTexture.h
```diff
@@ -2,35 +2,105 @@
...
     // Takes ownership of the image
-    VkTexture(const TextureDescription& description, ::vk::UniqueImage rawImage);
-    VkTexture(TextureDescription&& description, ::vk::UniqueImage rawImage);
+    VkImageTexture(const TextureDescription& description, ::vk::UniqueImage rawImage);
+    VkImageTexture(TextureDescription&& description, ::vk::UniqueImage rawImage);
 
-    // Creates a new image
-    // ...
-    VkTexture(VkGPUContext& ctx, TextureDescription&& description);
+    // Creates a new image from the description
```
Narvin-Chana, line 90: Having to subclass this makes me sad :(
Eet iz what eet iz.

examples/example_hello_triangle/HelloTriangleApplication.cpp
```diff
@@ -69,24 +69,14 @@ void HelloTriangleApplication::Run()
...
-                { static_cast<uint32_t>(width) / 8, static_cast<uint32_t>(height) / 8, 1 });
+            ctx.Dispatch({ .path = std::filesystem::current_path().parent_path().parent_path() / "examples" /
+                                   "example_hello_triangle" / "HelloTriangleShader.cs.hlsl",
+                           .entryPoint = "CSMain",
+                           .type = vex::ShaderType::ComputeShader },
+                         {},
+                         {},
+                         { { vex::ResourceBinding{ .name = "OutputTexture", .texture = workingTexture } } },
+                         { static_cast<uint32_t>(width) / 8, static_cast<uint32_t>(height) / 8, 1 });
             ctx.Copy(workingTexture, graphics->GetCurrentBackBuffer());
```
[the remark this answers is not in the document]
alexandreLBarrett: I honestly thought it was a merge mistake. I'll add them back

src/Vex/ShaderCompiler.cpp
```diff
@@ -223,18 +222,18 @@ std::expected<void, std::string> ShaderCache::CompileShader(RHIShader& shader)
...
+        //     shaderCompilationResults->GetOutput(DXC_OUT_REFLECTION, IID_PPV_ARGS(&reflectionBlob), nullptr);
+        // if (FAILED(reflectionResult))
+        // {
+        //     return std::unexpected("Unable to get the shader reflection blob after compilation.");
+        // }
+        //
+        // DxcBuffer reflectionBuffer;
+        // reflectionBuffer.Encoding = DXC_CP_ACP;
+        // reflectionBuffer.Ptr = reflectionBlob->GetBufferPointer();
+        // reflectionBuffer.Size = reflectionBlob->GetBufferSize();
```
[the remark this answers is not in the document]
alexandreLBarrett: Yeah. I say we do this in another MR since it will need a more thorough refactor of the ShaderCompiler. I think doing a way to generate reflection only would be my prefered solution

src/Vulkan/VkCommandList.cpp
```diff
@@ -79,59 +100,305 @@ void VkCommandList::SetPipelineState(const RHIGraphicsPipelineState& graphicsPip
...
 {
-    VEX_NOT_YET_IMPLEMENTED();
+    auto& vkPSO = reinterpret_cast<const VkComputePipelineState&>(computePipelineState);
+    commandBuffer->bindPipeline(CommandQueueTypeToPipelineBindPoint(type), *vkPSO.computePipeline);
 }
 
 void VkCommandList::SetLayout(RHIResourceLayout& layout)
 {
-    VEX_NOT_YET_IMPLEMENTED();
+    // nothing to do here i think
```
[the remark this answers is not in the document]
alexandreLBarrett: We should have a discussion on the API offered by CommandList now that i have something that works. There a couple of things that diverge from DX12 and I find some naming somewhat confusing

src/Vulkan/VkCommandList.cpp
alexandreLBarrett: std::unintialised_copy is the c++ typed equivalent of memcpy. It's just more cplusplusy

src/Vulkan/VkCommandList.cpp
alexandreLBarrett: I wrote this before understanding the layout transfer thing. You transfer layout. I think that the access and stage masks define on what to wait to change the layout. I think you could wait to be able to write to it without changing layout. I'm not sure though

src/Vulkan/VkRHI.cpp
alexandreLBarrett: actually since its a aggregate struct we can omit the second arg and it works just fine

src/Vulkan/VkTexture.cpp
alexandreLBarrett: I should double check that but there doesn't seem to be another one that is appropriate

src/Vex/Hash.h
```diff
@@ -21,16 +21,13 @@
 
 // Macro to generate hash function for a struct or class
 #define VEX_MAKE_HASHABLE(type, fields)                                                                                \
-    namespace std                                                                                                      \
-    {                                                                                                                  \
     template <>                                                                                                        \
-    struct hash<type>                                                                                                  \
+    struct std::hash<type>                                                                                             \
```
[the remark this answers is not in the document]
alexandreLBarrett: gcc in release doesnt like ::std for some reason... simply std it will be

src/Vulkan/VkCommandList.cpp
alexandreLBarrett: I added compute flag in case but this needs to be tested properly

src/Vulkan/VkTexture.cpp
alexandreLBarrett: General means all access types to this resource is allowed so yeah that would be the usecase for UAV

src/Vex/Bindings.h
```diff
@@ -20,7 +20,7 @@ struct ConstantBinding
     }
 
     void* data;
-    size_t size;
+    u32 size;
```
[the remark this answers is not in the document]
alexandreLBarrett: Lets make that change when we have global constants

src/Vulkan/VkCommandList.cpp
```diff
@@ -42,6 +48,21 @@ void VkCommandList::Close()
...
+{
+    switch (type)
+    {
+    case CommandQueueTypes::Compute:
+        return ::vk::PipelineBindPoint::eCompute;
+    // TODO: distinguish properly compute bind points from queue type. Cant bind a Compute PSO on a graphics bindpoint
+    case CommandQueueTypes::Graphics:
+        return ::vk::PipelineBindPoint::eCompute;
+    default:
+        VEX_ASSERT("No bind point found for command queues");
```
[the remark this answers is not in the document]
alexandreLBarrett: i missed a false in the first arg. The intent was to crash in default case. Assert or log fatal will have the same effect

src/Vulkan/VkCommandList.cpp
Narvin-Chana: Really? I thought it's purpose was more for if you want to copy data to a buffer of a certain type without calling the destructor on the elements of the buffer (thus uninitialised_copy).

src/Vex/RHI/RHIDescriptorPool.h
```diff
@@ -8,25 +10,7 @@
...
-    } type = Type::None;
-
-    static BindlessHandle CreateHandle(u32 index, u8 generation, Type type)
-    {
-        BindlessHandle handle = Handle::CreateHandle(index, generation);
-        handle.type = type;
-        return handle;
-    }
-};
+using BindlessHandle = Handle<>;
```
Narvin-Chana, line 13: The reason I didn't use usings for the handle system was to allow for compile time validation on handle types (eg not being able to pass a BindlessHandle to a function taking in another handle type). Does this maintain this functionality?

src/Vulkan/VkCommandList.cpp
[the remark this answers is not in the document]
alexandreLBarrett: Whe discussed moving more logic in the ResourceLayout class to be bound all at once

Narvin-Chana (approved):
LGTM!

PR #23, opened 2025-07-09
Resource Binding Shader Gen

Narvin-Chana:
- Contains a simple way to map resources to bindless indices passed via root/push constants automatically.
- Supposes that no local constants exist.
- TBD whether this scales well with multiple file includes but it is a first step.
- Users are expected to use `VEX_GLOBAL_RESOURCE(type, name);` in their shaders to declare the global resources that are used. 
- The C++ name of the resource should match the HLSL name.


alexandreLBarrett:
I noticed an issue with the push constants on the VK side. To access push constants on HLSL we need to add the [[vk::push_constant]] attribute in front of the constant buffer that will have the data. So you could add another define that is only defined on vulkan that you prepend to the constantbuffer used for bindless indices

alexandreLBarrett (approved):

PR #24, opened 2025-07-11
Added a way for ResourceCleanup to correctly free a texture's allocated bindless handles

Narvin-Chana:
- For now done with a simple std::visit, could potentially look into doing something more robust (some sort of OnPreDestroy method that all types registered with the ResourceCleanup would have).

### DIFF
--- a/src/DX12/DX12Texture.cpp
+++ b/src/DX12/DX12Texture.cpp
@@ -4,6 +4,7 @@
 
 #include <Vex/Logger.h>
 #include <Vex/Platform/Windows/WString.h>
+#include <Vex/RHI/RHIDescriptorPool.h>
 
 #include <DX12/DX12Formats.h>
 #include <DX12/HRChecker.h>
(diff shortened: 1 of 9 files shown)


src/DX12/DX12Texture.cpp
```diff
@@ -293,6 +294,17 @@ DX12Texture::~DX12Texture()
 {
 }
 
+void DX12Texture::FreeBindlessHandles(RHIDescriptorPool& descriptorPool)
+{
+    for (auto& [heapSlot, bindlessHandle] : cache | std::views::values)
```
alexandreLBarrett, line 299: You would probably need to clear the cache after no? or else you'll have dangling data

Narvin-Chana: Good point, the function is supposed to be called upon destruction, but its probably still a good idea to clear the cache in case we end up wanting to call this function in another context.

alexandreLBarrett (approved):

alexandreLBarrett (approved):

alexandreLBarrett (approved):

PR #29, opened 2025-07-25
Forced Graphics API to be selected at compile time

Narvin-Chana:
- Also refactored the cmake file to make it a bit more digestible
- ⏰ There's currently an issue where when using the "Clang | Development | Vulkan" preset, the application will crash upon reaching the `::vk::createInstanceUnique(instanceCI);` function call. The crash originates from deep inside Vk code, I'm afraid its some sort of memory corruption, potentially related to the cmake changes? I'll investigate further later on.⏰

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -18,63 +18,129 @@
       }
     },
     {
-      "name": "x64-MSVC-Base",
+      "name": "debug-build",
       "hidden": true,
-      "inherits": "base",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug"
+      }
+    },
+    {
+      "name": "development-build",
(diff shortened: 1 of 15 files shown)


CMakePresets.json
```diff
@@ -18,63 +18,129 @@
...
+    },
+    {
+      "name": "clang-shipping-dx12",
+      "displayName": "Clang | Release | DirectX 12",
+      "inherits": [ "base", "clang-compiler", "shipping-build", "dx12-api" ],
+      "condition": {
+        "type": "equals",
+        "lhs": "${hostSystemName}",
+        "rhs": "Windows"
+      }
```
alexandreLBarrett, line 138: pretty nice

PR #30, opened 2025-07-26
Added HLSL202x flags to the ShaderCompiler

Narvin-Chana:
These flags allow users to use the newest language features in HLSL. As per the [Microsoft devblog release](https://devblogs.microsoft.com/directx/dxc-1-8-2405-available/), I've combined the `-HV 202x` flag with: `-Wconversion -Wdouble-promotion -Whlsl-legacy-literal` in order to avoid users using future deprecated features.

As always this is completely optional for users of Vex, although this is opt-out, so by default this will be active.

Additionally:
- Changed how the user passes shader compiler settings to the shader compiler.
- Renamed example folders to no longer have a superfluous "example_" in their names, a future rename should be to specify "compute" on the original hello triangle, but I wanted to minimize the changes for currently open PRs.


src/Vex/PipelineStateCache.cpp
```diff
@@ -97,9 +97,9 @@ const RHIComputePipelineState* PipelineStateCache::GetComputePipelineState(const
     return ps;
 }
 
-ShaderCache& PipelineStateCache::GetShaderCache()
+ShaderCompiler& PipelineStateCache::GetShaderCache()
```
alexandreLBarrett, line 100: nit: get shader cache that returns ShaderCompiler seems odd

src/Vex/ShaderCompiler.h
```diff
@@ -82,6 +83,10 @@ struct ShaderCache
     void FlushCompilationErrors();
 
 private:
+    static constexpr auto HLSL202xFlags = {
+        L"-HV 202x", L"-Wconversion", L"-Wdouble-promotion", L"-Whlsl-legacy-literal"
```
alexandreLBarrett, line 87: Will this have issues with vulkan compatibility? in the case that some features are not transferable to SPIR-V

src/Vex/ShaderCompilerSettings.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+namespace vex
+{
+
+struct ShaderCompilerSettings
+{
+    // Determines if shaders should be compiled with debug symbols.
+    // Defaults to true in non-shipping builds and false in shipping.
+    bool enableShaderDebugging = !VEX_SHIPPING;
```
alexandreLBarrett, line 10: we should probably make it so its impossible to set this to true in shipping build?

alexandreLBarrett (approved):

PR #31, opened 2025-07-26
Removed VEX_HASH_COMBINE_ENUM in favor of adding enum support to the default VEX_HASH_COMBINE

Narvin-Chana:
- Also made the heart slightly bigger.

### DIFF
--- a/.github/workflows/cmake-multi-platform.yml
+++ b/.github/workflows/cmake-multi-platform.yml
@@ -12,21 +12,14 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     strategy:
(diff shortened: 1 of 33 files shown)


src/Vex/Hash.h
```diff
@@ -2,15 +2,30 @@
 
 #include <magic_enum/magic_enum.hpp>
 
+namespace vex
+{
+
+template <typename T>
+constexpr auto PurifyHashValue(const T& obj) -> decltype(auto)
```
alexandreLBarrett, line 9: `decltype(auto)` ? wouldn't just `auto` work?

PR #33, opened 2025-07-26
Changed command list submission to be batched as per vendor recommendations

Narvin-Chana:
Vendors recommend that command lists be submitted in one batch, this is for multiple reasons:
- Reduces driver overhead.
- Allows the driver to potentially concatenate the command lists together (if synchronization allows) which can lead to better performance.

Other changes:
- Also made DX12 RHI crash if single buffering is used, since it technically does not support it (instead of the previous way of handling things which was to still allocate two backbuffers and then try to lie that its supported).
- The Vulkan side of things was a bit more complicated than I expected, so I just kept the current logic (making sure no crashes occur). Added a trello to improve this: https://trello.com/c/J4WjMkVk

This PR also fixes a few issues with the GitHub actions yml (for example, clang-windows was actually not using clang at all).

### DIFF
--- a/src/DX12/DX12SwapChain.cpp
+++ b/src/DX12/DX12SwapChain.cpp
@@ -90,7 +90,7 @@ UniqueHandle<RHITexture> DX12SwapChain::CreateBackBuffer(u8 backBufferIndex)
 
 u8 DX12SwapChain::GetBackBufferCount(FrameBuffering frameBuffering)
 {
-    return std::max<u8>(2, std::to_underlying(frameBuffering));
+    return std::min(2, std::to_underlying(frameBuffering));
 }
 
 } // namespace vex::dx12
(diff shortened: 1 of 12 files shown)


src/DX12/DX12SwapChain.cpp
```diff
@@ -90,7 +90,13 @@ UniqueHandle<RHITexture> DX12SwapChain::CreateBackBuffer(u8 backBufferIndex)
...
 u8 DX12SwapChain::GetBackBufferCount(FrameBuffering frameBuffering)
 {
-    return std::max<u8>(2, std::to_underlying(frameBuffering));
+    if (frameBuffering == FrameBuffering::Single)
+    {
+        VEX_LOG(Fatal,
+                "Error creating swapchain: DX12 limits us to a minimum of 2 backbuffers, meaning single-buffering is "
+                "impossible.");
+    }
+    return std::to_underlying(frameBuffering);
```
alexandreLBarrett, line 99: This is probably the case in vk too lol

alexandreLBarrett (approved):

PR #36, opened 2025-07-27
DX12 Buffers

Narvin-Chana:
# This is dependent on #28 being merged first!
In this PR, I implement the DX12 side of buffers, while also reworking certain aspects of buffer creation to better match how we create textures. I've taken many creative liberties here, feel free to let me know if you disagree with anything!
I've also made sure the Vulkan side still works (and fixed the issues I've caused).

Features:
- Changed buffer description in order to allow for structured buffer usage on DX12 (on Vulkan it remains unsupported until further notice). This is accomplished via the "stride" field in the BufferDescription, non-zero values indicate that the buffer is structured, otherwise all buffers are considered as raw buffers (to be used with ByteAddressBuffer) in hlsl shaders.
- Changed buffer usage flags to allow for multiple usages (flags were previously exclusive), so for example you could have a buffer with simultaneous `shader resource` and `vertex buffer` usages, allowing for it to be bound to either (one at a time, of course).
- Simplified DX12 bindless descriptor creation, finally there's no use in storing a CPUHeap for UAV/CBV/SRV descriptors, instead we can just allocate a static descriptor and just create the view inside it directly. This greatly simplifies the code!
- Changed .clang-format to order includes by group according to the Vex convention which will avoid further hassle in PRs
- (mostly) fixed resizing in the HelloTriangle compute example, one crash remains on the vulkan side due to a synchronization (/semaphore) issue, I didn't have the courage to attack this issue, so I left it as-is for now (only occurs when resizing, so not the end of the world for now).


src/Vex/ResourceBindingSet.cpp
```diff
@@ -8,15 +8,16 @@ namespace vex
 
 void ResourceBindingSet::ValidateBindings() const
 {
-    ResourceBinding::ValidateResourceBindings(reads, ResourceUsage::Read);
-    ResourceBinding::ValidateResourceBindings(writes, ResourceUsage::UnorderedAccess);
+    ResourceBinding::ValidateResourceBindings(reads, TextureUsage::Read);
+    ResourceBinding::ValidateResourceBindings(writes, TextureUsage::UnorderedAccess);
```
alexandreLBarrett, line 12: Buffer usage could be added here for further validation on the buffers

src/Vulkan/VkCommandList.cpp
```diff
@@ -1,10 +1,10 @@
 ﻿#include "VkCommandList.h"
 
+#include <numeric>
```
alexandreLBarrett, line 3: unused?

src/Vulkan/VkBuffer.cpp
```diff
@@ -1,5 +1,7 @@
 ﻿#include "VkBuffer.h"
 
+#include <magic_enum/magic_enum.hpp>
```
alexandreLBarrett, line 3: include order? this seems to be considered as a std header maybe?

examples/example_hello_triangle/HelloTriangleShader.cs.hlsl
```diff
@@ -26,8 +34,12 @@ float sdf(float2 p)
     // Convert pixel coordinates to normalized space for opengl-based sdf function (-1 to 1)
     float2 uv = float2(dtid.xy) / max(width, height).xx * 2 - 1;
 
+#if defined(VEX_DX12)
+    float4 color = ColorBuffer[0].cols;
+#elif defined(VEX_VULKAN)
     float4 color = ColorBuffer.Load<float4>(0);
```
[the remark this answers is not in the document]
Narvin-Chana: A few thoughts about this:

In this situation its not at all about forcing the user to have separate macros, on the CPP side I've ifdef-ed the code to make the buffer structured on the DX12 side only. So this ifdef exists because both sides are using different binding logic, in the case the user is using raw buffers, the hlsl is identical.

Now in the case the user wants to use StructuredBuffers, then they are choosing to lean in heavily on the DX12 side, this in my opinion is not the end of the world as long as the user is aware that their choice has consequences on the usability of other APIs. Right now this choice isn't being correctly communicated, so potentially making this clearer would be a good idea, maybe via a log warning when using structured buffers on DX12 notifying that this renders their shaders potentially incompatible with Vulkan?
The root issue is that Vulkan does not\* support hlsl structured buffers with our binding model, and I'm not sure that this means that we should remove/hide the feature from the DX12 api side where it is compatible.

Next, about having a macro abstraction for this, I believe that shader code structures should match native HLSL as much as possible, I wouldn't want to force users to use custom load/store functions we provide, as this would reduce portability of Vex code to other hlsl renderers. Because of this, I believe that we should only use macros to provide native hlsl structures (in this case StructuredBuffer and/or ByteAddressBuffer) so that users familiar with hlsl are easily able to write code.

\* Now... that was a lie, I've recently figured out how to get structured buffers (and even constant buffers!) working in Vulkan HLSL. This can be accomplished (slightly painfully) by exploiting the fact that resource descriptor arrays can overlap. The idea is that  we use reflection before shader compilation to determine which structures the user would request bindlessly, and then create a ResourceDescriptorArray for that specific type, so something like this:

```hlsl
// Codegen: 
StructuredBuffer<MyStruct> MyStruct_StructuredBuffer[] : register(REG);
StructuredBuffer<MyStruct2> MyStruct2_StructuredBuffer[] : register(REG);

// User code:
struct MyStruct { /* ... */};
VEX_RESOURCE(StructuredBuffer<MyStruct>, MyBuffer);
struct MyStruct2 { /* ... */};
VEX_RESOURCE(StructuredBuffer<MyStruct2>, MyBuffer2);
```

Now the thing (you've probably noticed) that makes this VERY painful, is that we can't just insert the codegen at the beginning, as the type passed in doesn't yet exist. Instead we are forced to determine the first occurrence of each VEX_RESOURCE invocation with structured buffers, and place the corresponding structured buffer array just before it (and this only for the first instance).

I don't suggest doing this immediately, however this would allow for typed StructuredBuffers and ConstantBuffers to be used in our shaders (currently DX12 loses out on these because of vulkan hlsl restrictions)! Which imo would be really cool and remove the final restriction\*\* we've found with bindless.

\*\* Now... that was another lie, the pain factor increases tenfold when I mention that we would also have to support Structured Buffers/Constant Buffers being used inside a function scope, if the user manipulates bindless handles themselves. This would then require inserting the codegen just above the function, as declaring a resource array inside a function scope is not possible (I believe).

examples/ExampleApplication.h
```diff
@@ -1,9 +1,9 @@
 #pragma once
 
-#include <string_view>
-
 #include <Vex.h>
 
+#include <string_view>
```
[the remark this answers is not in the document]
Narvin-Chana: Yep I'll have to check the regex

src/Vex/ResourceBindingSet.cpp
Narvin-Chana: Will do

PR #40, opened 2025-07-31
Vulkan now supports HLSL's ResourceDescriptorHeap

Narvin-Chana:
- After asking for help on the Graphics Programmer Discord, I was told: 

> DXC has -fvk-bind-resource-heap <binding> <set> and -fvk-bind-sampler-heap <binding> <set> which you can use with mutable descriptors

So after investigating I added the ResourceHeap flag and reworked vulkan binding logic to use mutable descriptors. To my surprise this worked perfectly, greatly reducing the complexity of our shader gen!

There is a small runtime cost, but one that I think is worth it for ease of use.


alexandreLBarrett (commented):
Since the descriptor handling is now very similar if not identical on each API we could share the handle management?

alexandreLBarrett (approved):

PR #44, opened 2025-08-03
Support of all buffer types

alexandreLBarrett:
This adds the support for all buffer types, readonly (ConstantBuffer, StructuredBuffer, ByteAddressBuffer) and readwrite (RWStructuredBuffer, RWByteAddressBuffer). This is done by separating the Usages possible with the buffer and how the buffer will be used when binding it to a shader.

I also added validation of the buffer desc at creation. It validates the name and the byte size for now.

I also moved the stride information to the ResourceBinding because its only need when creating the views on the buffers. This has a side effect of needing the specify the stride in the resource binding, but I think it makes sense if the user wants to use the same buffer for different purposes.

### DIFF
--- a/src/DX12/RHI/DX12Buffer.h
+++ b/src/DX12/RHI/DX12Buffer.h
@@ -12,6 +12,29 @@ namespace vex
 {
 struct BufferDescription;
 }
+
+namespace vex::dx12
+{
+
+struct BufferViewCacheKey
+{
+    BufferBindingUsage usage;
+    u32 stride;
+
+    bool operator==(const BufferViewCacheKey& other) const = default;
+};
+
+} // namespace vex::dx12
+
+// clang-format off
+
+VEX_MAKE_HASHABLE(vex::dx12::BufferViewCacheKey,
+    VEX_HASH_COMBINE(seed, obj.usage);
+    VEX_HASH_COMBINE(seed, obj.stride);
+);
+
+// clang-format on
+
 namespace vex::dx12
 {
 
@@ -23,7 +46,7 @@ public:
     virtual std::span<u8> Map() override;
     virtual void Unmap() override;
 
-    virtual BindlessHandle GetOrCreateBindlessView(BufferUsage::Type usage, RHIDescriptorPool& descriptorPool) override;
+    virtual BindlessHandle GetOrCreateBindlessView(BufferBindingUsage usage, u32 stride, RHIDescriptorPool& descriptorPool) override;
     virtual void FreeBindlessHandles(RHIDescriptorPool& descriptorPool) override;
 
     ID3D12Resource* GetRawBuffer()
@@ -38,7 +61,7 @@ private:
 
     ComPtr<ID3D12Resource> buffer;
 
-    std::unordered_map<BufferUsage::Type, BindlessHandle> viewCache;
+    std::unordered_map<BufferViewCacheKey, BindlessHandle> viewCache;
 };
 
 } // namespace vex::dx12
--- a/src/Vex/Buffer.h
+++ b/src/Vex/Buffer.h
@@ -4,6 +4,8 @@
 
 #include <Vex/EnumFlags.h>
 #include <Vex/Handle.h>
+#include <Vex/Logger.h>
+#include <Vex/Resource.h>
 #include <Vex/Types.h>
 
 namespace vex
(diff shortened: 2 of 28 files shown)


src/Vex/Buffer.h
```diff
@@ -13,33 +15,71 @@ namespace vex
...
+};
 
-    bool IsStructured() const
+inline void ValidateBufferDescription(const BufferDescription& desc)
+{
+    if (desc.name.empty())
     {
-        return stride != 0;
+        VEX_LOG(Fatal, "The buffer needs a name on creation.");
     }
```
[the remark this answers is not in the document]
alexandreLBarrett: This isn't as important because this only for debug and logging purposes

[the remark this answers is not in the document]
Narvin-Chana: Oh yeah, I didn't see this was the buffer creation, not binding nvm then.

alexandreLBarrett:
I'll do it next as well as the resource binding API. I feel like it could be streamlined a lot.

src/DX12/RHI/DX12Buffer.h
```diff
@@ -12,6 +12,22 @@ namespace vex
...
+{
+    BufferBindingUsage usage;
+    u32 stride;
+
+    bool operator==(const BufferViewCacheKey& other) const = default;
+};
+
+} // namespace vex::dx12
+
+VEX_MAKE_HASHABLE(vex::dx12::BufferViewCacheKey, VEX_HASH_COMBINE(seed, obj.usage); VEX_HASH_COMBINE(seed, obj.usage););
```
Narvin-Chana, line 29: hash combine is hashing the usage twice. You probably forgot the stride.

src/Vex/Buffer.h
```diff
@@ -13,33 +14,76 @@ namespace vex
...
-    CPUVisible                      = 1 << 6, // Can be read by CPU
-    CPUWrite                        = 1 << 7, // Can be written to by CPU
-    // GPU visiblity is deduced from other flags.
+    GenericBuffer                   = 1 << 0, // Buffers that can be read from shaders (SRV)
+    UniformBuffer                   = 1 << 1, // Buffers with specific alignment constraints (CBV)
+    ReadWriteBuffer                 = 1 << 2, // Buffers with read and write operations in shaders (UAV)
+    VertexBuffer                    = 1 << 3, // Buffers used for vertex buffers
+    IndexBuffer                     = 1 << 4, // Buffers used for index buffers
+    IndirectArgs                    = 1 << 5,
+    RaytracingAccelerationStructure = 1 << 6,
```
Narvin-Chana, line 23: Should generic be the 0 value? right now generic = 1
Since every buffer is by default generic, it makes more sense imo to have it be the 0 value.

src/Vex/Buffer.h
```diff
@@ -13,33 +14,76 @@ namespace vex
...
 END_VEX_ENUM_FLAGS();
 
+// Defines what the specific binding will bind as
+// maps directly to the type that will be used in the
+// shader to access the buffer
+enum class BufferBindingUsage
```
Narvin-Chana, line 29:
```suggestion
enum class BufferBindingUsage : u8
```

src/Vex/Buffer.h
```diff
@@ -13,33 +14,76 @@ namespace vex
...
+        return usages & BufferUsage::ReadWriteBuffer;
+    }
+
+    return false;
+}
+
+enum class ResourceMemoryLocality : u8
+{
+    GPUOnly, CPURead, CPUWrite,
+};
```
Narvin-Chana, line 62: This should ideally be in Resource.h

Narvin-Chana (changes requested):
Some small fixes/questions but we're almost there!

src/Vex/Buffer.h
alexandreLBarrett: can't really because 0 isn't a bit flag value. It can't be tested against.

alexandreLBarrett: Tested doesn't work

Narvin-Chana: Discussed together on discord, should keep this as-is while making sure that CreateBuffer checks that at least one usage flag is set.

src/Vulkan/RHI/VkBuffer.cpp
```diff
@@ -42,7 +42,7 @@ static ::vk::BufferUsageFlags GetVkBufferUsageFromDesc(const BufferDescription&
 
     if (!flags)
     {
-        VEX_LOG(Fatal, "Buffer usage not supported by VulkanRHI.");
+        flags |= eTransferSrc;
```
Narvin-Chana, line 45: Shouldn't this always be true? In case for example we want to duplicate a buffer?

Narvin-Chana (approved):
Nice 👍

PR #45, opened 2025-08-07
Graphics pipeline on vulkan

alexandreLBarrett:


src/Vulkan/RHI/VkCommandList.cpp
```diff
@@ -388,8 +475,71 @@ void VkCommandList::Transition(std::span<std::pair<RHIBuffer&, RHIBufferState::F
...
 }
 
+void VkCommandList::BeginRendering(const RHIDrawResources& resources)
+{
+    ::vk::Rect2D maxArea{ { 0, 0 }, { 0, 0 } };
+    for (auto& renderTargets : resources.renderTargets)
+    {
+        maxArea.extent.width = std::max(renderTargets.texture->GetDescription().width, maxArea.extent.width);
+        maxArea.extent.height = std::max(renderTargets.texture->GetDescription().height, maxArea.extent.height);
+    }
```
alexandreLBarrett, line 485: is this necessary or we can assume that all render targets will have the same size?

src/Vulkan/RHI/VkTexture.cpp
```diff
@@ -9,6 +9,8 @@
 #include <Vulkan/VkGPUContext.h>
 #include <Vulkan/VkMemory.h>
 
+#include "Vex/RHIBindings.h"
```
alexandreLBarrett, line 12: header formatting

src/Vulkan/RHI/VkTexture.h
```diff
@@ -1,5 +1,7 @@
 ﻿#pragma once
 
+#include "VkMacros.h"
```
alexandreLBarrett, line 3: include format

src/DX12/RHI/DX12Texture.cpp
```diff
@@ -388,6 +388,22 @@ void DX12Texture::FreeBindlessHandles(RHIDescriptorPool& descriptorPool)
...
+
+    if (description.usage & TextureUsage::DepthStencil)
+    {
+        return RHITextureState::DepthWrite;
+    }
+
+    VEX_LOG(Fatal,
+            "Invalid texture passed to ClearTexture, your texture must allow for either RenderTarget usage or "
+            "DepthStencil usage.");
+}
```
Narvin-Chana, line 406: So in this case you prioritise the RenderTarget over DepthStencil usage?
Maybe it would be better to return ::Flags instead? SOmething like GetValidClearTextureStates()
Also haven't finished the review, but it seems like this shouldn't be API-specific?

src/RHI/RHICommandList.h
```diff
@@ -54,6 +56,10 @@ class RHICommandListInterface
     virtual void Transition(std::span<std::pair<RHITexture&, RHITextureState::Flags>> textureNewStatePairs) = 0;
     virtual void Transition(std::span<std::pair<RHIBuffer&, RHIBufferState::Flags>> bufferNewStatePairs) = 0;
 
+    // Need to be called before and after all draw commands with the same DrawBinding
+    virtual void BeginRendering(const RHIDrawResources& resources) = 0;
+    virtual void EndRendering() = 0;
```
Narvin-Chana, line 61: This is probably a great idea, we could potentially change the entire CommandContext draw/dispatch structure to reproduce something similar to this? Would avoid having to cache PSOs and various other states etc...
It makes for a more optimized (and lean) abstraction as well.

src/Vex/Containers/DirtyFlagged.cpp
```diff
@@ -0,0 +1,5 @@
+﻿#include "DirtyFlagged.h"
+
+namespace vex
+{
+} // namespace vex
```
Narvin-Chana, line 5: .cpp probably not needed.

src/Vex/Containers/DirtyFlagged.h
```diff
@@ -0,0 +1,82 @@
+﻿#pragma once
+#include <utility>
+
+namespace vex
+{
+
+template <class T>
+class DirtyFlagged
+{
+    mutable bool dirtyFlag = true;
```
Narvin-Chana, line 10: We can keep it if you see a practical future use for it. If its just theoretical for now, I'd maybe keep it in a stash somewhere?

src/Vex/Containers/DirtyFlagged.h
```diff
@@ -0,0 +1,82 @@
+﻿#pragma once
+#include <utility>
+
+namespace vex
+{
+
+template <class T>
```
Narvin-Chana, line 7: You probably want to enforce that T is default constructible?

src/Vex/Containers/StdContainers.h
```diff
@@ -0,0 +1,19 @@
...
+struct U32Alloc : std::allocator<T>
+{
+    using size_type = u32;
+};
+
+// This allows us to use u32 as size values and indices
+template <class T>
+using vector = std::vector<T, U32Alloc<T>>;
+
+} // namespace vex
```
Narvin-Chana, line 19: How safe is this? I wasn't able to figure out if changing the size_type of the allocator has unintended consequences.
Also would use `Vector` to match our naming convention.

src/Vex/GraphicsPipeline.h
```diff
@@ -142,6 +142,7 @@ struct DepthStencilState
     bool depthWriteEnabled = false;
     CompareOp depthCompareOp = CompareOp::None;
     bool depthBoundsTestEnabled = false; // Vulkan only
+    bool stencilTestEnabled = false;     // Vulkan only
```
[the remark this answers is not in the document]
Narvin-Chana: Yeah it could be used in `DX12GraphicsPipeline.cpp` line 165:
```cpp
    // This:
    desc.StencilEnable = (depthStencilState.front.compareOp != CompareOp::Always ||
                          depthStencilState.back.compareOp != CompareOp::Always);
    // Becomes:
    desc.StencilEnable = depthStencilState.stencilTestEnabled;
```
I didn't specify it since I supposed we could deduce stencil usage from the compare ops used, but its probably cleaner to have a separate flag :)

src/Vulkan/RHI/VkCommandList.cpp
```diff
@@ -10,6 +10,9 @@
 #include <Vulkan/RHI/VkTexture.h>
 #include <Vulkan/VkErrorHandler.h>
 
+#include "Vex/Containers/StdContainers.h"
+#include "Vex/DrawHelpers.h"
```
Narvin-Chana, line 14:
```suggestion
#include <Vex/Containers/StdContainers.h>
#include <Vex/DrawHelpers.h>
```

src/Vulkan/RHI/VkCommandList.cpp
```diff
@@ -187,14 +190,76 @@ void VkCommandList::SetDescriptorPool(RHIDescriptorPool& descriptorPool, RHIReso
...
+    //         },
+    //         .baseArrayLayer = clearBinding.startSlice,
+    //         .layerCount = layerCount,
+    //     };
+    //
+    //     commandBuffer->clearAttachments(1, &clearAttachment, 1, &clearRect);
+    // }
+    // else
+
+    if (desc.usage & TextureUsage::DepthStencil &&
```
Narvin-Chana, line 231: Sick! This code is way cleaner than the DX side thanks to ImageSubresourceRange!
Commented code is for clearing inside a clear rect? We probably should have a task to add ClearRects support on the API level (I've added a Trello).

src/Vulkan/RHI/VkCommandList.cpp
```diff
@@ -244,6 +309,10 @@ ::vk::ImageMemoryBarrier2 GetMemoryBarrierFrom(VkTexture& texture, RHITextureSta
         barrier.srcAccessMask = AccessFlagBits2::eMemoryRead | AccessFlagBits2::eMemoryWrite;
         barrier.srcStageMask = PipelineStageFlagBits2::eAllCommands;
         break;
+    case ImageLayout::eColorAttachmentOptimal:
+        barrier.dstAccessMask = AccessFlagBits2::eShaderWrite;
+        barrier.dstStageMask = PipelineStageFlagBits2::eAllGraphics | PipelineStageFlagBits2::eComputeShader;
+        break;
     default:
         VEX_ASSERT(false, "Transition source image layout not supported");
     }
```
Narvin-Chana, line 318: In this function I've noticed that the prevLayout sometimes changes dstAccessMask and dstStageMask instead of the src**, is this intended? I've fixed it in my synchronization/constants branch since it was causing issues with incorrect layouts.

src/Vulkan/RHI/VkCommandList.cpp
Narvin-Chana: So having variable size render targets is supposed to work, so iterating on the RTs is valid.
Where I'm a bit confused is about what value the renderArea should have (min or max of all RTs?).
The vk docs are not very clear imo: https://registry.khronos.org/vulkan/specs/latest/man/html/VkRenderPassBeginInfo.html#_description

src/Vulkan/RHI/VkCommandList.cpp
```diff
@@ -388,8 +457,71 @@ void VkCommandList::Transition(std::span<std::pair<RHIBuffer&, RHIBufferState::F
...
+    };
+
+    const ::vk::RenderingInfo info{
+        .renderArea = maxArea,
+        .layerCount = 1,
+        .viewMask = 0,
+        .colorAttachmentCount = static_cast<uint32_t>(colorAttachmentsInfo.size()),
+        .pColorAttachments = colorAttachmentsInfo.data(),
+        .pDepthAttachment = depthInfo ? &*depthInfo : nullptr,
+        .pStencilAttachment = depthInfo ? &*depthInfo : nullptr,
```
Narvin-Chana, line 498: Smart pointers would really benefit from a get_address_of...

src/Vulkan/RHI/VkPipelineState.cpp
```diff
@@ -8,11 +10,68 @@
 #include <Vulkan/RHI/VkTexture.h>
 #include <Vulkan/VkErrorHandler.h>
 
+#include "../VkMacros.h"
+#include "Vulkan/VkFormats.h"
```
Narvin-Chana, line 14: includes yabadibaba

src/Vulkan/RHI/VkPipelineState.cpp
```diff
@@ -8,11 +10,68 @@
...
+
 namespace vex::vk
 {
 
-VkGraphicsPipelineState::VkGraphicsPipelineState(const Key& key)
+namespace GraphicsPiplineUtils
+{
+
+VEX_VK_BEGIN_ENUM_MAPPING(InputTopology, InputTopology, ::vk::PrimitiveTopology, VkTopology)
```
Narvin-Chana, line 22: Mmmmmmmmmacros!

src/Vulkan/RHI/VkPipelineState.cpp
```diff
@@ -8,11 +10,68 @@
...
+VEX_VK_ENUM_MAPPING_ENTRY(CounterClockwise, eCounterClockwise)
+VEX_VK_END_ENUM_MAPPING
+
+VEX_VK_BEGIN_ENUM_MAPPING_STATIC(PolygonMode, PolygonMode, ::vk::PolygonMode, VkPolygonMode)
+VEX_VK_ENUM_MAPPING_ENTRY(Fill, eFill)
+VEX_VK_ENUM_MAPPING_ENTRY(Line, eLine)
+VEX_VK_ENUM_MAPPING_ENTRY(Point, ePoint)
+VEX_VK_END_ENUM_MAPPING
+
+} // namespace GraphicsPiplineUtils
```
Narvin-Chana, line 50: Could (should?) move this into VkGraphicsPipeline.h/.cpp to match Vex/DX12 file structure.

src/Vulkan/RHI/VkPipelineState.cpp
```diff
@@ -22,12 +81,174 @@ void VkGraphicsPipelineState::Compile(const Shader& vertexShader,
...
+    };
+
+    ::vk::PipelineViewportStateCreateInfo viewportStateCI{
+        .viewportCount = 0,
+        .pViewports = nullptr,
+        .scissorCount = 0,
+        .pScissors = nullptr,
+    };
+
+    ::vk::GraphicsPipelineCreateInfo graphicsPipelineCI{ .pNext = &pipelineRenderingCI,
```
Narvin-Chana, line 227: I didn't check in detail this section since its a LOT of boilerplate conversion/setup code.
The main things to know are that invalid states should be checked at some point on our side (I believe the DX12-side has a validation function inside CreateGraphicsPipelineState in its RHI, something similar for the most egregious errors could be nice.
Also MSAA is unsupported (so any attempt to use it should error out).

src/Vulkan/RHI/VkRHI.cpp
```diff
@@ -305,7 +306,7 @@ UniqueHandle<RHICommandPool> VkRHI::CreateCommandPool()
 
 UniqueHandle<RHIGraphicsPipelineState> VkRHI::CreateGraphicsPipelineState(const GraphicsPipelineStateKey& key)
 {
-    return MakeUnique<VkGraphicsPipelineState>(key);
+    return MakeUnique<VkGraphicsPipelineState>(key, *device, *PSOCache);
 }
```
Narvin-Chana, line 310: I wonder if we couldn't just declare these on the stack? Now that we have no need for polymorphism and considering that PSO pointers are NEVER stored outside of the PSO cache itself (only used then discarded).

Narvin-Chana: I guess it might make resource cleanup a bit annoying though.

src/Vulkan/RHI/VkTexture.h
```diff
@@ -7,7 +7,12 @@
 
 #include <Vulkan/RHI/VkDescriptorPool.h>
 #include <Vulkan/VkHeaders.h>
+#include <Vulkan/VkMacros.h>
```
Narvin-Chana, line 10: vk macros unused?

Narvin-Chana (changes requested):
Good stuff! We're on a roll rn ye ye
Didn't clone locally but I'm guessing you tested all examples with both APIs.

src/Vulkan/RHI/VkRHI.cpp
alexandreLBarrett: We could store them in the PSO cache unordered map

src/DX12/RHI/DX12Texture.cpp
alexandreLBarrett: This should return the state the texture needs to be to be cleared. Technically you can always use general in Vulkan and clear with that. If I use RenderTarget it needs to be on an image that is bound to rendering at that point. So the logic is a bit different from DX12

src/Vex/PipelineStateCache.cpp
```diff
@@ -32,12 +32,12 @@ const RHIGraphicsPipelineState* PipelineStateCache::GetGraphicsPipelineState(con
     RHIGraphicsPipelineState* ps;
     if (graphicsPSCache.contains(key))
     {
-        ps = graphicsPSCache[key].get();
+        ps = &graphicsPSCache.at(key);
```
Narvin-Chana, line 35: having used contains, you can probably just use [], no?

src/Vex/PipelineStateCache.cpp
```diff
@@ -32,12 +32,12 @@ const RHIGraphicsPipelineState* PipelineStateCache::GetGraphicsPipelineState(con
     RHIGraphicsPipelineState* ps;
     if (graphicsPSCache.contains(key))
```
Narvin-Chana, line 33: Now that I think about it, this should be `find`, then either dereference the find result or create and insert.

src/Vex/PipelineStateCache.cpp
```diff
@@ -71,12 +71,12 @@ const RHIComputePipelineState* PipelineStateCache::GetComputePipelineState(const
     RHIComputePipelineState* ps;
     if (computePSCache.contains(key))
     {
-        ps = computePSCache[key].get();
+        ps = &computePSCache.at(key);
```
Narvin-Chana, line 74: Same as previous comments

Narvin-Chana (commented):
Just a few things, but this is almost there!

src/DX12/RHI/DX12Texture.cpp
Narvin-Chana: But the state requested depends on the binding (/clear color) of the user, no? Can a texture be DepthStencil and RenderTarget at the same time? Probably not, if they indeed can't in both DX12 and Vulkan then I'd add validation upon texture creation, and keep this code as-is.

alexandreLBarrett: I don't think they can considering they need to be in different layouts to be used in the graphics pipeline

Narvin-Chana (approved):

PR #53, opened 2025-08-17
Fixed Vulkan resizing and fullscreen

alexandreLBarrett:
Surface capabilities were never refreshed and the swapchain therefore never had the right size. Which caused an out of date present.


Narvin-Chana (approved):
LGTM!

Narvin-Chana (changes requested):
Actually just remembered the following:
currentBackBufferID of VkSwapChain is unused, this is probably not correct.
Please investigate and see if its no longer needed.

PR #54, opened 2025-08-17
Vulkan debug object naming

alexandreLBarrett:
Added debug names on most objects. Let me know if more objects need naming

### DIFF
--- a/src/Vulkan/VkDebug.h
+++ b/src/Vulkan/VkDebug.h
@@ -1,5 +1,6 @@
 ﻿#pragma once
 
+#include <Vulkan/VkErrorHandler.h>
 #include <Vulkan/VkHeaders.h>
 
 namespace vex::vk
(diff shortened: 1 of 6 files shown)


src/Vulkan/VkDebug.h
```diff
@@ -10,4 +11,20 @@ VKAPI_ATTR VkBool32 VKAPI_CALL debugCallback(::vk::DebugUtilsMessageSeverityFlag
                                              const ::vk::DebugUtilsMessengerCallbackDataEXT* pCallbackData,
                                              void* pUserData);
 
-}
\ No newline at end of file
+#if defined(VEX_DEBUG) or defined(VEX_DEVELOPMENT)
```
Narvin-Chana, line 14: FYI they are always defined, so it should just be `#if VEX_DEBUG or VEX_DEVELOPMENT`. (or `#if !VEX_SHIPPING` to match DX12's debug naming).

Narvin-Chana (changes requested):
Just one defined() issue, otherwise all good!

PR #55, opened 2025-08-17
Ray Tracing Vex+DX12 Implementation

Narvin-Chana:
- Adds in the whole pipeline to compile and use ray tracing shaders in DX12.
- Adds the abstraction in Vex for performing ray tracing using DXR (vulkan's RT is stubbed out currently).
- Adds a new example with a simple rayGen shader invocation to write in a texture.


src/DX12/RHI/DX12Buffer.cpp
```diff
@@ -197,4 +198,37 @@ void DX12Buffer::FreeBindlessHandles(RHIDescriptorPool& descriptorPool)
...
+
+    // Ensures proper alignment - shader records must be aligned to D3D12_RAYTRACING_SHADER_RECORD_BYTE_ALIGNMENT (32
+    // bytes)
+    u64 alignedRecordStride = AlignUp<u64>(recordStride, D3D12_RAYTRACING_SHADER_RECORD_BYTE_ALIGNMENT);
+    shaderTableStride = alignedRecordStride;
+
+    std::span<u8> mappedData = Map();
+
+    // Write each shader identifier.
+    for (u64 i = 0; i < shaderIdentifiers.size(); ++i)
```
alexandreLBarrett, line 217: nit: Couldn't you align shaderIdentifier elements to 32 bits in memory. This would allow for a simple memcpy

src/DX12/RHI/DX12Buffer.h
```diff
@@ -46,14 +46,30 @@ class DX12Buffer final : public RHIBufferBase
...
 
-    virtual BindlessHandle GetOrCreateBindlessView(BufferBindingUsage usage, u32 stride, RHIDescriptorPool& descriptorPool) override;
+    virtual BindlessHandle GetOrCreateBindlessView(BufferBindingUsage usage,
+                                                   u32 stride,
+                                                   RHIDescriptorPool& descriptorPool) override;
     virtual void FreeBindlessHandles(RHIDescriptorPool& descriptorPool) override;
 
+    void WriteShaderTable(std::span<void*> shaderIdentifiers,
+                          u64 shaderIdentifierSize = D3D12_SHADER_IDENTIFIER_SIZE_IN_BYTES,
+                          u64 recordStride = D3D12_SHADER_IDENTIFIER_SIZE_IN_BYTES);
```
alexandreLBarrett, line 56: Does this need to be a member function? I feel like this is not the buffer's responsibility

src/Vex/CommandContext.h
```diff
@@ -23,6 +23,7 @@ struct TextureClearValue;
 struct DrawDescription;
 struct DrawResources;
 struct ComputeResources;
+struct RayTracingPassDescription;
```
alexandreLBarrett, line 26: Not sure if I like the "Pass" term. We don't really use it anywhere. We should be consistent

src/Vex/PipelineStateCache.h
```diff
@@ -34,10 +35,18 @@ class PipelineStateCache
...
     ShaderCompiler& GetShaderCompiler();
 
 private:
+    // Converts a RayTracingPSOKey into the mirrored version of itself which contains all required shaders.
+    std::optional<RayTracingShaderCollection> GetRayTracingShaderCollection(const RHIRayTracingPipelineState::Key& key,
+                                                                            ShaderResourceContext resourceContext);
+
+    // Can't use NonNullPtr, as this field can potentially be empty during the time it takes for the RHI to be
+    // initialized.
     RHI* rhi;
```
alexandreLBarrett, line 50: Lets add a `LikelyNotNullPtr` then ;)

src/Vex/CommandContext.h
Narvin-Chana: Couldn't be named _resources, because it doesn't include the resources...
Any suggestions? This only contains the shaders the user wants to use in their RT pass.

src/DX12/RHI/DX12Buffer.h
Narvin-Chana: In DX12/DXR, shader tables ARE buffers, so in this case this is simply filling in our buffer as a ShaderTable. I think this is a valid DX12-specific usage. Maybe I could rename the method to SetDataAsShaderTable or something similar for clarity?

src/DX12/RHI/DX12Buffer.cpp
Narvin-Chana: Each void* in the span points towards memory located elsewhere. Adding padding to the void* vector would not achieve the purpose of copying the contents the void* points to, into the mappedData.

eg:
```cpp
T1 a = ...;
T2 b = ...;
vector<void*> identifiers = { &a, &b };

// Structure of the final buffer:
vector finalBuffer = { contents of a, padding to 32 bits, contents of b, padding to 32 bits };
```
It might be possible to do this in 1 memcpy, but it would probably end up being more complicated. We'd have to, when obtaining the shader identifiers, dereference them, and copy them into a buffer.

I personally think its not worth it vs this which works well enough (and is only called once per RT shader type used in a RT PSO).

src/DX12/RHI/DX12Buffer.h
alexandreLBarrett: To me, this is more of a specific use of the buffer than an intrinsic property of the buffer. The buffer behaves the same way just has different data in it. I would make a utility function that fills the buffer with the correct data according to your RT needs. Maybe I'm misunderstanding the concept but just looking at the code. It doesn't seem to change the way the buffer behaves

Narvin-Chana: The only issue is that we have to store data relating to the shader table, most notably the alignment, I could maybe move this outside of buffer into a free function.

alexandreLBarrett: To me this falls in a simllar bucket as the stride. We dont store it in the buffer itself. could handled in a similar way? Maybe we should store the stride and alignment in the buffer but this constraints the buffer to a single usage whichi is something I think we intended to leave somewhat flexible

Narvin-Chana: Good point, moved into a new wrapper class that just uses a buffer and stores the stride. This makes more sense!

alexandreLBarrett (approved):

PR #59, opened 2025-08-19
Various Fixes following integration of latest version of Vex in Zest (my renderer)

Narvin-Chana:
- Bindings no longer fail in ClearTexture and in GetBindlessHandle (texture and buffer) if you don't pass in a handle (and if you don't pass in a specific binding usage).
- The depth stencil validation no longer errors out on valid state.


examples/hello_triangle_graphics_pipeline/HelloTriangleGraphicsApplication.cpp
```diff
@@ -68,7 +68,7 @@ void HelloTriangleGraphicsApplication::Run()
             ctx.ClearTexture(vex::TextureBinding{
                 .name = "Backbuffer",
                 .texture = graphics->GetCurrentBackBuffer(),
-            }, vex::TextureUsage::RenderTarget, clearValue);
+            }, clearValue);
```
alexandreLBarrett, line 71: Are you going to make the change to use TextureRange instead? or later?

alexandreLBarrett (approved):

PR #60, opened 2025-08-22
Shader compiler modularization

Narvin-Chana:
Separates DXC from our shader compiler's more general logic, in anticipation of integrating Slang as a shader compiler backend.

The choice between shader compilers is made at backend creation during runtime, and this is implemented via pimpl, tbd if we want this choice to be made at compile time.


src/Vex/Shaders/DXCImpl.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+static std::vector<DxcDefine> ConvertDefinesToDxcDefine(const std::vector<ShaderDefine>& defines)
+{
+    std::vector<DxcDefine> dxcDefines;
+    dxcDefines.reserve(defines.size());
+    for (auto& d : defines)
+    {
+        dxcDefines.emplace_back(DxcDefine{ .Name = d.name.c_str(), .Value = d.value.c_str() });
+    }
+    return dxcDefines;
```
alexandreLBarrett, line 68: nit: could use std::transform here

src/Vex/Shaders/DXCImpl.cpp
```diff
@@ -0,0 +1,304 @@
...
+    shaderSource.Ptr = shaderBlob->GetBufferPointer();
+    shaderSource.Size = shaderBlob->GetBufferSize();
+    shaderSource.Encoding = DXC_CP_ACP; // Assume BOM says UTF8 or UTF16 or this is ANSI text.
+
+    std::vector<LPCWSTR> args;
+    args.reserve(shaderEnv.args.size());
+    for (const std::wstring& arg : shaderEnv.args)
+    {
+        args.emplace_back(arg.c_str());
+    }
```
alexandreLBarrett, line 153: nit: could use transform here as well

alexandreLBarrett (approved):

src/Vex/Shaders/DXCImpl.cpp
Narvin-Chana: changed

PR #61, opened 2025-08-25
Rework of Binding Logic (Part 1 of Slang Integration)

Narvin-Chana:
## Binding logic
Vex shaders no longer explicitly have to specify global resources, instead we generate the bindings automatically on the C++ side, since we already have all the required information.
The gives us the following shader code:
```cpp
// old_file.hlsl
VEX_SHADER
{
    VEX_LOCAL_CONSTANTS(
        MyStruct,
        MyLocalConstants
    );
    VEX_GLOBAL_RESOURCE(ConstantBuffer<Colors>, ColorBuffer);
}
// new_file.hlsl
VEX_SHADER_BINDINGS
```

In the DXCImpl, we define `VEX_SHADER_BINDINGS` to define the bindless resources as was done previously with `VEX_SHADER`, but also to place and declare the global resources automatically. The typing information (eg: `Texture2D<float3> myTex`) is deduced from A: The texture type, B: the new `typeName` field in TextureBinding ( C++). and C: the name in the TextureBinding (C++).

As an example for the shader above we auto-generate the following for DX12:
```cpp
struct Vex_GeneratedGlobalResources
{
    uint ColorBuffer_BindlessIndex; 
}; 
ConstantBuffer<Vex_GeneratedGlobalResources> Vex_GeneratedGlobalResourcesCB : register(b0);
ConstantBuffer<MyStruct> LocalConstants : register(b1);
static ConstantBuffer<Colors> ColorBuffer = ResourceDescriptorHeap[Vex_GeneratedGlobalResourcesCB.ColorBuffer_BindlessIndex];
```

And Vulkan:
```cpp
struct Vex_GeneratedGlobalResources
{ 
    uint ColorBuffer_BindlessIndex;
};
struct Vex_GeneratedCombinedResources
{    
    uint GlobalResourcesBindlessIndex;
    MyStruct UserData;
};
[[vk::push_constant]] ConstantBuffer<Vex_GeneratedCombinedResources> Vex_GeneratedCombinedResourcesCB;
static ConstantBuffer<Vex_GeneratedGlobalResources> Vex_GeneratedGlobalResourcesCB = ResourceDescriptorHeap[Vex_GeneratedCombinedResourcesCB.GlobalResourcesBindlessIndex];
static ConstantBuffer<Colors> ColorBuffer = ResourceDescriptorHeap[Vex_GeneratedGlobalResourcesCB.ColorBuffer_BindlessIndex];
```

This IMO has multiple advantages:
1. No duplicate information. Previously we had identical information in C++ and in HLSL that served just the purpose of complicating code-gen.
2. Avoids the ordering/sorting issue, we can just use the C++ order directly for resource injection.
3. Changing bindings in ANY case requires a recompilation and restarting the C++ program. So this adds 0 overhead to iteration times.
4. The code required for generating bindings is VERY simple versus what we had before. This makes for more rigid/stable codegen and is way more performant.
5. An issue the previous codegen had, was that line numbers became misaligned! Since we just define the macro with a one line definition errors messages now point to the correct line number!

There are some disadvantages though:
1. The user must now use the constantly-named "LocalConstants" variable to access their local constants. No more custom name for the constants. If this is a dealbreaker, it would be possible to pass it in on the C++ side. I didn't implement this since I wanted your feedback on this in general.
2. ConstantBinding now also require the type name to be passed in on the C++ side. This _could_ be avoided if we just use a constant struct name (eg: forcing the user to always use `VexLocalConstants`, or something similar).
3. Same thing for TextureBinding and BufferBinding, we have to request a new typeName field for codegen.

## Other changes
I have temporarily removed/disabled "smart" shader recompilation (avoiding recompiling unchanged shaders) due to the fact that the pre-processed hash solution won't work for Slang. Instead I think this should be tackled later on separately due to the complexity of resolving includes manually/caching various files' last_modified_time. This is a pretty isolated task so if it interests you you can attack this problem 😉

Added Slang to the project. Currently the SlangCompilerImpl is stubbed out, and disabled. Will be tackled in a future PR, separate to this one. I decided to separate them since this PR was already getting pretty big.


cmake/VexSlang.cmake
```diff
@@ -0,0 +1,74 @@
...
+        #"${slang_SOURCE_DIR}/bin/slang-llvm.dll"
+        "${slang_SOURCE_DIR}/bin/slang-rt.dll"
+        "${slang_SOURCE_DIR}/bin/slang.dll"
+    )
+elseif(UNIX)
+    set(SLANG_SHARED_LIBS 
+        "${slang_SOURCE_DIR}/lib/libgfx.so"
+        # These are potentially not necessary, for now left as disabled.
+        #"${slang_SOURCE_DIR}/lib/libslang-glsl-module.dll"
+        #"${slang_SOURCE_DIR}/lib/libslang-glslang.dll"
```
alexandreLBarrett, line 41: These comments can be removed on linux as .dlls are windows only

examples/hello_triangle/HelloTriangleApplication.cpp
```diff
@@ -112,13 +112,15 @@ void HelloTriangleApplication::Run()
...
-                        .usage = vex::BufferBindingUsage::ConstantBuffer
+                        .usage = vex::BufferBindingUsage::ConstantBuffer,
+                        .typeName = "Colors",
                     },
                     vex::BufferBinding{
                         .name = "CommBuffer",
                         .buffer = commBuffer,
                         .usage = vex::BufferBindingUsage::RWStructuredBuffer,
-                        .stride = sizeof(float) * 4
+                        .typeName = "float4",
```
alexandreLBarrett, line 122: Would rename it to "resourceValueType" maybe I feel like typename is ambiguous

examples/hello_triangle/HelloTriangleShader2.cs.hlsl
```diff
@@ -1,9 +1,5 @@
-VEX_SHADER
-{
-    VEX_GLOBAL_RESOURCE(RWTexture2D<float4>, OutputTexture);
-    VEX_GLOBAL_RESOURCE(Texture2D<float4>, SourceTexture);
-    VEX_GLOBAL_RESOURCE(StructuredBuffer<float4>, CommBuffer);
-}
+// Must also be declared before any code uses the bound global resources.
+VEX_SHADER_BINDINGS
```
alexandreLBarrett, line 2: Would you consider using a different format for any codegen macros? Like `[(VEX_BINDINGS)]` (this is just an example) as this would make it obvious its code gen and has specific properties. Wdyt?

examples/hello_triangle_graphics_pipeline/HelloTriangleGraphicsShader.hlsl
```diff
@@ -58,7 +51,7 @@ float4 PSMain(VSOutput input)
     : SV_Target
 {
     static const float moveSpeed = 0.2f;
-    float offset = sin(MyLocalConstants.time * moveSpeed) * 0.1;
+    float offset = sin(LocalConstants.time * moveSpeed) * 0.1;
```
alexandreLBarrett, line 54: Nit : Could use Uniforms instead of LocalConstants maybe? (Less chars ;) )

src/Vex/Shaders/DXCImpl.h
```diff
@@ -29,31 +29,27 @@ using ComPtr = CComPtr<T>;
...
 namespace vex
 {
 
-struct DXCCompilerImpl : public CompilerInterface
+struct ShaderDefine;
+
+struct DXCCompilerImpl : public CompilerBase
 {
-    DXCCompilerImpl();
+    DXCCompilerImpl(std::vector<std::filesystem::path> includeDirectories);
```
alexandreLBarrett, line 41: Const& here?

src/Vex/Bindings.h
```diff
@@ -16,13 +17,20 @@ struct ConstantBinding
...
-        : data{ static_cast<const void*>(&data) }
+    ConstantBinding(std::string typeName, const T& data)
+        : typeName(std::move(typeName))
+        , data{ static_cast<const void*>(&data) }
         , size{ sizeof(T) }
     {
     }
 
+    // Shader struct type to use for the constant binding.
+    std::string typeName;
```
alexandreLBarrett, line 28: Couldn add the binding name to let the user specify it instead of LocalConstants this would make it more coherent with other bindings

src/Vex/Bindings.h
```diff
@@ -55,11 +65,16 @@ struct BufferBinding
...
     std::string name;
     // The texture to bind
     Texture texture;
+    // The usage of the texture.
     TextureBindingUsage usage = TextureBindingUsage::None;
+    // The underlying shader type to use for the texture.
+    // Only valid HLSL/Slang primitives are accepted.
+    // Not passing a texture typename will default to float4.
+    // eg: "float4" for an RGBA texture, "uint" for a R8_UINT texture, etc...
+    std::string typeName = "float4";
```
alexandreLBarrett, line 77: Classic code injection entry point lol. Buuuut, were not web devs so we dont bother with this

src/Vex/Shaders/DXCImpl.h
Narvin-Chana: It's STD moved into the parent which itself std moves into the member field.

src/Vex/Bindings.h
Narvin-Chana: Ye lol, I think shaders are already pretty easy to hack into.

cmake/VexSlang.cmake
Narvin-Chana: There are Linux .so for this I just made a mistake when copy pasting. Will fix

Narvin-Chana:
Will be reopened with code from another branch following all the things we've discussed.

PR #62, opened 2025-08-27
VkAllocator and staging buffer rework

alexandreLBarrett:
Implemented the VKAllocator to allow the new memory allocation in VK as well. 

Moved all of the staging code from the buffer class to the graphics backend as its not really the buffer's responsibility to have its staging data. The RAII object is provided to the user to write its data in it. Given the command context it will abstract the staging buffer creation and uploading for the user. This could cause issues if the user holds the RAII object beyond the context's life but its the simplest i could find without rewriting a lot of logic. We could force to provide the data and do it all at once for this API. It's the user's responsibility to do a staging buffer if it doesnt go through the one-time API.

I added a couple of stubs to allow for the next step of uploading data to textures.

I'll tackle the uploading to textures next


src/Vex/CommandContext.cpp
```diff
@@ -401,6 +375,11 @@ void CommandContext::Copy(const Buffer& source, const Buffer& destination)
     cmdList->Copy(sourceRHI, destinationRHI);
 }
 
+void CommandContext::Copy(const Buffer& source, const Texture& destination)
+{
+    VEX_NOT_YET_IMPLEMENTED();
+}
+
```
Narvin-Chana, line 382: For uploading data to textures we will probably want this to be the default path (using staging buffer instead of staging textures). This allows for less waste, as buffers are generally more tightly packed versus textures which can have padding and such, but renders the upload a bit more tricky due to alignment constraints.

src/Vex/Resource.cpp
```diff
@@ -0,0 +1,67 @@
...
+{
+    VEX_ASSERT(mappedData.size() >= inData.size());
+    // This function is defined in <utility> which is more lightweight than <ranges>/<algorithm>.
+    std::ranges::copy(inData, mappedData.begin());
+}
+
+void ResourceMappedMemory::SetData(std::span<u8> inData)
+{
+    SetData(std::span<const u8>{ inData.begin(), inData.end() });
+}
```
Narvin-Chana, line 65: This is very probably deprecated if we keep resources mapped when possible.

src/DX12/RHI/DX12Buffer.h
```diff
@@ -46,15 +45,15 @@ class DX12Buffer final : public RHIBufferBase
...
-    virtual void Unmap() override;
-
     virtual BindlessHandle GetOrCreateBindlessView(BufferBindingUsage usage,
                                                    u32 stride,
                                                    RHIDescriptorPool& descriptorPool) override;
     virtual void FreeBindlessHandles(RHIDescriptorPool& descriptorPool) override;
     virtual void FreeAllocation(RHIAllocator& allocator) override;
 
+    std::span<u8> Map() override;
+    void Unmap() override;
```
Narvin-Chana, line 55: Code consistency
```suggestion
    virtual std::span<u8> Map() override;
    virtual void Unmap() override;
```

src/Vex/GfxBackend.cpp
```diff
@@ -218,12 +218,89 @@ Buffer GfxBackend::CreateBuffer(BufferDescription description, ResourceLifetime
...
+
+    Buffer stagingBuffer =
+        CreateBuffer(GetStagingBufferDescription(texture.description.name, texture.description.GetTextureByteSize()),
+                     ResourceLifetime::Static);
+    RHIBuffer& rhiStagingBuffer = GetRHIBuffer(stagingBuffer.handle);
+    std::move_only_function<void()> transitionAndUpload = [this,
+                                                           stagingBuffer,
+                                                           &cmdList = ctx.GetRHICommandList(),
+                                                           &rhiDestTexture = GetRHITexture(texture.handle),
+                                                           &rhiStagingBuffer = rhiStagingBuffer] mutable
```
Narvin-Chana, line 295: mmm that's a beefy lambda with many (potentially risky) captures

src/Vex/Texture.cpp
```diff
@@ -72,4 +72,10 @@ void ValidateTextureDescription(const TextureDescription& description)
 
 } // namespace TextureUtil
 
+u32 TextureDescription::GetTextureByteSize() const
+{
+    // TODO: Implement this correctly...
+    // Very generous and not at all accurate. Especially with mips
+    return width * height * depthOrArraySize * (4 * sizeof(float));
+}
```
Narvin-Chana, line 80: This should be api specific. Texture sizes depend on platforms.

src/Vulkan/RHI/VkAllocator.cpp
```diff
@@ -1,16 +1,127 @@
...
+    }
+
+    VEX_LOG(Fatal, "Unsuitable memory found for flags {:x}", static_cast<u32>(flags))
+    return 0;
+}
+
+} // namespace AllocatorUtils
+
+VkAllocator::VkAllocator(NonNullPtr<VkGPUContext> ctx)
+    : RHIAllocatorBase(ctx->physDevice.getMemoryProperties().memoryTypeCount)
```
Narvin-Chana, line 55: I'm guessing the reason that the memory type count is now dynamic (and no longer correlated to the HeapType enum) is due to vulkan rigidity versus potentially weird hardware?

src/Vulkan/RHI/VkAllocator.cpp
```diff
@@ -1,16 +1,127 @@
...
+{
+    memoryPagesByType.resize(pageInfos.size());
+}
+
+VkAllocator::~VkAllocator()
+{
+    for (u32 i = 0; i < pageInfos.size(); ++i)
+    {
+        for (auto& [_, value] : memoryPagesByType[i])
+            ctx->device.freeMemory(value);
```
Narvin-Chana, line 66: There's no RAII way to hold and thus clear-on-destroy, memory? That sucks!

src/Vulkan/RHI/VkAllocator.h
```diff
@@ -1,18 +1,51 @@
...
+    ::vk::DeviceMemory GetMemoryFromAllocation(const Allocation& allocation);
 
 protected:
-    virtual void OnPageAllocated(PageHandle handle, HeapType heapType) override;
-    virtual void OnPageFreed(PageHandle handle, HeapType heapType) override;
+    virtual void OnPageAllocated(PageHandle handle, u32 memoryTypeIndex) override;
+    virtual void OnPageFreed(PageHandle handle, u32 memoryTypeIndex) override;
+
+    // Not using UniqueDeviceMemory here because of wierd quirk with maps in vectors
+    std::vector<std::unordered_map<PageHandle, ::vk::DeviceMemory>> memoryPagesByType;
```
Narvin-Chana, line 46: Ah! Did the piecewise_construct not work? Seems a bit sucky to not be able to keep uniformity with other vk code.

src/Vulkan/RHI/VkBuffer.cpp
```diff
@@ -1,14 +1,14 @@
-﻿#include "VkBuffer.h"
+﻿#include "VkAllocator.h"
 
 #include <magic_enum/magic_enum.hpp>
 
 #include <Vex/Buffer.h>
 
+#include <Vulkan/RHI/VkBuffer.h>
```
Narvin-Chana, line 7: Includes are incorrect (main include is allocator in buffer.cpp)

src/Vex/Resource.h
```diff
@@ -1,10 +1,21 @@
...
+#include <span>
+#include <variant>
+
 #include <Vex/Handle.h>
+#include <Vex/RHIFwd.h>
 #include <Vex/Types.h>
 
+#include "RHI/RHIAllocator.h"
+
+#define VEX_USE_CUSTOM_ALLOCATOR_BUFFERS 1
```
Narvin-Chana, line 14: Nit: would change this to be wrapped by #ifndef / #endif to allow for potential user override.

src/Vex/Resource.h
```diff
@@ -25,4 +36,49 @@ struct BindlessHandle : Handle<BindlessHandle>
...
+
+    template <class T>
+    void SetData(const T& inData);
+
+private:
+    std::span<u8> mappedData;
+
+    IMappableResource& resource;
+
+    bool mapped = false;
```
Narvin-Chana, line 72: Would rename to `isMapped` or some sort of equivalent term.

src/Vulkan/RHI/VkBuffer.cpp
```diff
@@ -83,14 +64,21 @@ VkBuffer::VkBuffer(NonNullPtr<VkGPUContext> ctx, const BufferDescription& desc)
 
     const ::vk::MemoryRequirements reqs = ctx->device.getBufferMemoryRequirements(*buffer);
 
+#if VEX_USE_CUSTOM_ALLOCATOR_BUFFERS
+    auto allocResult = allocator.AllocateResource(desc.memoryLocality, reqs);
```
Narvin-Chana, line 68: Nit: Could use a structured binding here to be more explicit than first, second.

src/Vulkan/RHI/VkTexture.cpp
```diff
@@ -1,15 +1,16 @@
-﻿#include "VkTexture.h"
+﻿#include "VkAllocator.h"
 
 #include <Vex/Bindings.h>
 #include <Vex/RHIBindings.h>
 
+#include <Vulkan/RHI/VkAllocator.h>
 #include <Vulkan/RHI/VkCommandPool.h>
 #include <Vulkan/RHI/VkDescriptorPool.h>
+#include <Vulkan/RHI/VkTexture.h>
```
Narvin-Chana, line 9: Same as buffer, some include incorrectness with "VkAllocator" being the main include.

src/Vulkan/RHI/VkTexture.cpp
```diff
@@ -290,18 +309,23 @@ void VkTexture::CreateImage()
 
     ::vk::MemoryRequirements imageMemoryReq = ctx->device.getImageMemoryRequirements(*imageTmp);
 
+#if VEX_USE_CUSTOM_ALLOCATOR_BUFFERS
+    auto allocResult = allocator.AllocateResource(description.memoryLocality, imageMemoryReq);
```
Narvin-Chana, line 313: Nit: Same as buffer, could use structured binding.

src/Vex/Texture.cpp
alexandreLBarrett: Yeah i'm not really sure how to tackle that properly. I need to look deeper into it. I'll fix that in the texture upload MR if thats ok with you

src/Vex/GfxBackend.cpp
alexandreLBarrett: This would disapear with the changes you mentionned above regarding staging code

src/Vex/GfxBackend.cpp
```diff
@@ -218,12 +218,89 @@ Buffer GfxBackend::CreateBuffer(BufferDescription description, ResourceLifetime
...
+    {
+        cmdList.Transition(rhiStagingBuffer, RHIBufferState::CopySource);
+        cmdList.Transition(rhiDestTexture, RHIBufferState::CopyDest);
+        cmdList.Copy(rhiStagingBuffer, rhiDestTexture);
+        // Delete staging buffer at the end of the frame
+        resourceCleanup.CleanupResource(bufferRegistry.ExtractElement(stagingBuffer.handle));
+    };
+    return ResourceMappedMemory(rhiStagingBuffer, std::move(transitionAndUpload));
```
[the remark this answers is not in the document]
alexandreLBarrett: Fair enough. I mentionned in the PR desc that we could do something simillar to that

src/Vex/Resource.cpp
alexandreLBarrett: Would remove in said mr. This was to keep something functional in main

src/Vex/CommandContext.cpp
alexandreLBarrett: To be done in next MR thi

src/Vulkan/RHI/VkAllocator.cpp
alexandreLBarrett: I could have some kind of wrapper to pass to the unordered map but this seemed dirty to me

src/Vex/CommandContext.cpp
Narvin-Chana: Yeah buffer/texture region would be nice . I'd still keep the simple version of this function though. For ease of use in the general case.

Narvin-Chana: And here for the specific buffer->texture restrictions for dx12: https://learn.microsoft.com/en-us/windows/win32/direct3d12/upload-and-readback-of-texture-data

src/RHI/RHITexture.h
```diff
@@ -6,6 +6,8 @@
 #include <Vex/Texture.h>
 #include <Vex/Types.h>
 
+#include <RHI/RHIAllocator.h>
```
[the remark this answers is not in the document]
alexandreLBarrett: Same as the buffe answer ;)

src/RHI/RHIBuffer.h
```diff
@@ -1,12 +1,11 @@
 #pragma once
 
-#include <span>
-
 #include <Vex/Buffer.h>
+#include <Vex/NonNullPtr.h>
 #include <Vex/RHIFwd.h>
+#include <Vex/RHIImpl/RHIAllocator.h>
```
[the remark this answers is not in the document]
Narvin-Chana: Couldn't allocation be moved into Allocation.h instead?

PR #63, opened 2025-08-28
Slang Integration and Code-Gen removal

Narvin-Chana:
Two big aspects of this PR, the first being the integration of Slang as an optional Shader Compiler. 
- Users can now use .slang shaders on top of .hlsl ones, with full feature parity. 
- The DXC compiler remains fully supported and feature complete.
- Users can use both compiler interchangeably at runtime allowing for comparisons.

The second feature is the complete removal of shader codegen:
- Vex now ships with a Vex.hlsli (or Vex.slang) file that is copied to the application's working directory, which is automatically added to the include paths and can be `#include` (or `import`)ed by the user in their shaders. This contains all helper code we provide for use with Vex shaders. 
- The HUGE advantage of this, is that users can now have correct syntax highlighting (as long as you correctly configure your IDE) which was impossible with the previous solutions as the shader code depended on C++ context.
- Users now manually handle bindless handles and transition the underlying resources manually (resources like RenderTargets or DepthStencil are still automatically bound).
- Some helpers could be provided to simplify the handling of this. For now I've done as best as I can to make the usage painless (or as painless as possible).


examples/hello_triangle/HelloTriangleApplication.cpp
```diff
@@ -72,99 +72,116 @@ void HelloTriangleApplication::Run()
...
-                {
-                    static_cast<vex::u32>(std::ceil(static_cast<float>(width) / 8.0f)),
-                    static_cast<vex::u32>(std::ceil(static_cast<float>(height) / 8.0f)),
-                    1
-                });
-            ctx.Copy(finalOutputTexture, graphics->GetCurrentBackBuffer());
+                    vex::ConstantBinding(pass2Handles.data(), pass2Handles.size() * sizeof(vex::BindlessHandle)),
+                    {
+                        static_cast<vex::u32>(std::ceil(static_cast<float>(width) / 8.0f)),
+                        static_cast<vex::u32>(std::ceil(static_cast<float>(height) / 8.0f)),
```
alexandreLBarrett, line 178: This line makes my eyes bleed. stupid conversion warnings

examples/hello_triangle/HelloTriangleApplication.cpp
```diff
@@ -72,99 +72,116 @@ void HelloTriangleApplication::Run()
...
+                        1,
+                    });
+#endif
+            }
 
-            // clang-format on
```
alexandreLBarrett, line 167: Sooooooo... its gonna be back to ugly format (: (I'm too lazy to fix the formatting file)

shaders/Vex.hlsli
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#if defined(VEX_VULKAN)
+#define VEX_UNIFORMS(type, name) [[vk::push_constant]] type name;
```
alexandreLBarrett, line 4: This works with a struct without using constant buffers?

src/DX12/RHI/DX12Buffer.cpp
```diff
@@ -142,8 +142,8 @@ BindlessHandle DX12Buffer::GetOrCreateBindlessView(BufferBindingUsage usage,
             srvDesc.ViewDimension = D3D12_SRV_DIMENSION_BUFFER;
             srvDesc.Shader4ComponentMapping = D3D12_DEFAULT_SHADER_4_COMPONENT_MAPPING;
             srvDesc.Buffer.FirstElement = 0;
-            srvDesc.Buffer.NumElements = desc.byteSize / stride;
-            srvDesc.Buffer.StructureByteStride = stride;
+            srvDesc.Buffer.NumElements = desc.byteSize / *stride;
```
alexandreLBarrett, line 145: should validate if stride has value before deref

src/RHI/RHIDescriptorPool.h
```diff
@@ -24,7 +24,7 @@ class RHIDescriptorPoolBase
     bool IsValid(BindlessHandle handle);
 
 protected:
-    static constexpr u32 DefaultPoolSize = 8192;
+    static constexpr u32 DefaultPoolSize = 65536;
```
alexandreLBarrett, line 27: magic numbers? where did you get that metric?

src/Vex/Shaders/DXCImpl.cpp
```diff
@@ -149,6 +105,10 @@ std::expected<std::vector<u8>, std::string> DXCCompilerImpl::CompileShader(
 
     std::vector<LPCWSTR> args;
     args.reserve(shaderEnv.args.size());
+    for (const std::wstring& arg : shaderEnv.args)
+    {
+        args.emplace_back(arg.c_str());
+    }
     std::ranges::transform(shaderEnv.args,
                            std::back_inserter(args),
                            [](const std::wstring& arg) { return arg.c_str(); });
```
alexandreLBarrett, line 114: duplicate code? loop first then transform that does the same

src/Vex/Shaders/SlangImpl.cpp
```diff
@@ -0,0 +1,253 @@
...
+        !program)
+    {
+        return std::unexpected("Unable to create composite component type.");
+    }
+
+    // TODO: this is where reflection would be done (using program)
+
+    Slang::ComPtr<slang::IComponentType> linkedProgram;
+    diagnostics = nullptr;
+    if (SLANG_FAILED(program->link(linkedProgram.writeRef(), diagnostics.writeRef())) || diagnostics)
```
alexandreLBarrett, line 109: `|| diagnostics` here seems to do nothing since its set to `nullptr` before.

src/Vex/Bindings.h
```diff
@@ -14,15 +15,23 @@ namespace vex
...
-        , size{ sizeof(T) }
+        : ConstantBinding(static_cast<const void*>(&data), sizeof(T))
     {
     }
 
+    // Constant binding data.
     const void* data;
+
+    // Byte size of the local constant binding.
     u32 size;
```
alexandreLBarrett, line 35: nit: could we use spans instead? since we dont own the resource anyway

src/Vex/CommandContext.cpp
```diff
@@ -401,6 +332,46 @@ void CommandContext::Copy(const Buffer& source, const Buffer& destination)
...
+void CommandContext::TransitionBindings(std::span<const ResourceBinding> resourceBindings)
+{
+    // Collect all underlying RHI textures.
+    std::vector<RHITextureBinding> rhiTextureBindings;
+    std::vector<RHIBufferBinding> rhiBufferBindings;
+    ResourceBindingUtils::CollectRHIResources(*backend, resourceBindings, rhiTextureBindings, rhiBufferBindings);
+
+    // This code will be greatly simplified when we add caching of transitions until the next GPU operation.
+    // See: https://trello.com/c/kJWhd2iu
+    CommandContext_Internal::TransitionAndCopyFromStaging(*cmdList, rhiBufferBindings);
```
alexandreLBarrett, line 370: dont forget to remove this after rebase

src/RHI/RHIDescriptorPool.h
Narvin-Chana: This tooootallly magic number is coincidentally 2^16 🙂

It gives a pool size of 2-4MB and is already a more sensible default than only allowing 1k resources.

src/DX12/RHI/DX12Buffer.cpp
Narvin-Chana: I'll double check that this validation is being done on bindings in ctx.GetBindlessHandles just in case.

shaders/Vex.hlsli
Narvin-Chana: Yep, isn't that cool?

examples/hello_triangle/HelloTriangleApplication.cpp
Narvin-Chana: Also with my changes we have less nesting now.

src/Vex/Bindings.h
Narvin-Chana: That's not a nit, but instead a very valid point.

alexandreLBarrett (approved):

PR #64, opened 2025-08-29
Added Indexed Drawing + Hello Cube example with Depth Texture, Vertex Buffer and Index Buffer

Narvin-Chana:
- Only contains the DX12 implementation for setting VBs/IBs, however the Vulkan part shouldn't be difficult to add.

### DIFF
--- a/src/Vex/CommandContext.h
+++ b/src/Vex/CommandContext.h
@@ -112,7 +112,7 @@ public:
     // Allows the passed in lambda to be executed in a draw scope.
     void ExecuteInDrawContext(std::span<const TextureBinding> renderTargets,
                               std::optional<const TextureBinding> depthStencil,
-                              std::function<void()> callback);
+                              const std::function<void()>& callback);
 
     // Returns the RHI command list associated with this context (you should avoid using this unless you know
     // what you are doing).
--- a/src/Vulkan/RHI/VkCommandList.cpp
+++ b/src/Vulkan/RHI/VkCommandList.cpp
@@ -541,16 +541,12 @@ void VkCommandList::DrawIndexed(
 
 void VkCommandList::SetVertexBuffers(u32 startSlot, std::span<RHIBufferBinding> vertexBuffers)
 {
-    VEX_NOT_YET_IMPLEMENTED();
-    // Code should be verified by our resident vulkan expert.
-    // Unsure if vkOffsets should be bytes or bits?
-    static constexpr u64 ByteToBits = 8;
     std::vector<::vk::Buffer> vkBuffers(vertexBuffers.size());
     std::vector<::vk::DeviceSize> vkOffsets(vkBuffers.size());
     for (auto& [binding, buffer] : vertexBuffers)
     {
         vkBuffers.emplace_back(buffer->GetNativeBuffer());
-        vkOffsets.push_back(binding.offsetByteSize.value_or(0) * ByteToBits);
+        vkOffsets.push_back(binding.offsetByteSize.value_or(0));
     }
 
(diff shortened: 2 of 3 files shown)


src/Vex/CommandContext.h
```diff
@@ -94,29 +96,35 @@ class CommandContext
...
+    // However, in the case you are leveraging bindless resources, you are responsible for ensuring any used resources
+    // are in the correct state. This contains redundancy checks so feel free to call it even if the resource is
+    // potentially already in the desired state for correctness.
     void Transition(const Buffer& buffer, RHIBufferState::Type newState);
 
+    // Useful for calling native API draws when wanting to render to a specific Render Target.
+    // Allows the passed in lambda to be executed in a draw scope.
+    void ExecuteInDrawContext(std::span<const TextureBinding> renderTargets,
+                              std::optional<const TextureBinding> depthStencil,
+                              std::function<void()> callback);
```
alexandreLBarrett, line 115: would use const& for callback here as we dont know what will be passed

src/Vex/CommandContext.h
```diff
@@ -94,29 +96,35 @@ class CommandContext
...
+
     // Returns the RHI command list associated with this context (you should avoid using this unless you know
     // what you are doing).
     RHICommandList& GetRHICommandList();
 
 private:
+    std::optional<RHIDrawResources> PrepareDrawCall(const DrawDescription& drawDesc,
+                                                    const DrawResourceBinding& drawBindings,
+                                                    std::optional<ConstantBinding> constants);
```
alexandreLBarrett, line 124: intentional copy here?

src/Vulkan/RHI/VkCommandList.cpp
```diff
@@ -516,7 +518,47 @@ void VkCommandList::Draw(u32 vertexCount)
...
+    commandBuffer->setViewportWithCount(1, &*cachedViewport);
+    commandBuffer->setScissorWithCount(1, &*cachedScissor);
+    commandBuffer->drawIndexed(indexCount, instanceCount, indexOffset, vertexOffset, instanceOffset);
+}
+
+void VkCommandList::SetVertexBuffers(u32 startSlot, std::span<RHIBufferBinding> vertexBuffers)
+{
+    VEX_NOT_YET_IMPLEMENTED();
+    // Code should be verified by our resident vulkan expert.
+    // Unsure if vkOffsets should be bytes or bits?
```
alexandreLBarrett, line 546: Most likely bytes. I don't see anything in the doc about a bit offset. It says its a memory offset which i would assume is bytes as you cant really address bits with an address.

Would use bytes for now pls

src/Vulkan/RHI/VkCommandList.cpp
```diff
@@ -516,7 +518,47 @@ void VkCommandList::Draw(u32 vertexCount)
...
+        vkBuffers.emplace_back(buffer->GetNativeBuffer());
+        vkOffsets.push_back(binding.offsetByteSize.value_or(0) * ByteToBits);
+    }
+
+    commandBuffer->bindVertexBuffers(startSlot, static_cast<u32>(vkBuffers.size()), vkBuffers.data(), vkOffsets.data());
+}
+
+void VkCommandList::SetIndexBuffer(const RHIBufferBinding& indexBuffer)
+{
+    VEX_NOT_YET_IMPLEMENTED();
```
alexandreLBarrett, line 561: I suppose there is an associated trello ticket?

src/Vex/CommandContext.h
Narvin-Chana: Nope, good catch

alexandreLBarrett (approved):

PR #68, opened 2025-09-06
Fixed invalid swapchain handling in vulkan

alexandreLBarrett:
If at any point the swapchain cannot be acquired. We set the backend in an "invalid swapchain" state which does not acquire swapchain or present. It still submits the deferred commandlists though. From the user's standpoint its commands are always submitted. It's just that nothing is presented. A resize set the swapchain as valid again

### DIFF
--- a/src/Tests/SynchronizationTorture.cpp
+++ b/src/Tests/SynchronizationTorture.cpp
@@ -229,59 +229,57 @@ void SynchronizationTortureTest(GfxBackend& graphics)
 
     // Test 6: Resource Upload Torture
     //
-    // Disabled since we don't yet have buffer->texture upload.
-    //
-    // VEX_LOG(Info, "Test 6: Resource Upload Torture");
-    //{
-    //    // Create upload buffer
-    //    BufferDescription uploadBufDesc{};
-    //    uploadBufDesc.name = "Test6 Buf";
-    //    uploadBufDesc.byteSize = 1024 * 1024; // 1MB
-    //    uploadBufDesc.usage = BufferUsage::None;
-    //    uploadBufDesc.memoryLocality = ResourceMemoryLocality::CPUWrite;
-    //    auto uploadBuffer = graphics.CreateBuffer(uploadBufDesc, ResourceLifetime::Static);
+    VEX_LOG(Info, "Test 6: Resource Upload Torture");
+    {
+        // Create upload buffer
+        BufferDescription uploadBufDesc{};
+        uploadBufDesc.name = "Test6 Buf";
+        uploadBufDesc.byteSize = 1024 * 1024; // 1MB
+        uploadBufDesc.usage = BufferUsage::None;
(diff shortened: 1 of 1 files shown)


src/Vex/GfxBackend.cpp
```diff
@@ -125,6 +125,13 @@ void GfxBackend::Present(bool isFullscreenMode)
         VEX_LOG(Fatal, "Cannot present without using a swapchain!");
     }
 
+    if (!isSwapchainValid)
```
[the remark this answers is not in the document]
Narvin-Chana: Also we should call CleanupResources()

src/Vex/GfxBackend.cpp
```diff
@@ -349,8 +362,9 @@ void GfxBackend::OnWindowResized(u32 newWidth, u32 newHeight)
 {
     // Do not resize if any of the dimensions is 0, or if the resize gives us the same window size as we have
     // currently.
-    if (newWidth == 0 || newHeight == 0 ||
-        (newWidth == description.platformWindow.width && newHeight == description.platformWindow.height))
+    if (isSwapchainValid &&
```
Narvin-Chana, line 365: Resizing with a size of 0 will cause a crash. We should only resize if:
The backbuffer is invalid AND newsize != 0.
And then only if backbuffer is invalid OR newSize != PrevSize.

src/Vulkan/RHI/VkSwapChain.cpp
```diff
@@ -152,9 +152,7 @@ RHITexture VkSwapChain::AcquireBackBuffer(u8 frameIndex)
                                                &currentBackbufferId);
     if (res == ::vk::Result::eErrorOutOfDateKHR)
     {
-        VEX_LOG(Error,
```
Narvin-Chana, line 155: Should we still output a warning?

Narvin-Chana: Vex should definitely continue working, I meant that we might still want to warn the user in the case they attempt to present while minimized. Via a log message.

alexandreLBarrett: This just happends on the first frame when minimised after that no frames are presented until resize. To me its implicit (or at least) has no impact that the frame is not presented so there would only be a warning once. It doesnt change the users code. The user can manage minimizing itself if needed but i wouldn't force it on people. Could add some verbose logging. 

Personnally i would expect the backend to work when i minimise my windows without having to manage a lot of code. If we want to force the user to manage it resize would need to be called with 0,0 when minimising. This way vex would know it shouldnt present. We could then say out of date is an error and force the user to manage properly its resizing. It's a design choice on how much we want to manage it or leave it to the user to do it.

Narvin-Chana: Ok I'm fine with keeping this under the hood then.

Narvin-Chana (approved):
LGTM!

PR #72, opened 2025-09-13
RHI Barrier Rework

Narvin-Chana:
- Changed synchro to use DX12's enhanced barriers.
- This allows us to use modern barriers everywhere, instead of faked states.


src/Vulkan/RHI/VkBarrier.cpp
```diff
@@ -0,0 +1,139 @@
...
+        return eMemoryRead;
+    case MemoryWrite:
+        return eMemoryWrite;
+    default:
+        VEX_LOG(Fatal, "Unsupported RHIBarrierAccess type.");
+        std::unreachable();
+    }
+}
+
+::vk::ImageLayout RHITextureLayoutToVulkan(RHITextureLayout textureLayout)
```
alexandreLBarrett, line 107: Could use the macro made for this but idm

alexandreLBarrett (approved):

alexandreLBarrett (approved):

PR #76, opened 2025-09-25
Various fixes following integration of Vex into my own renderer 

Narvin-Chana:
- Texture Sampler can now take in a non-mutable TextureSampler span
- Added helper to construct TextureSampler
- Fixed Vulkan not taking into account the entrypoint for PSO creation.
- Some additional error handling.
- Fixed multi-RTV output not working in DX12
- Fixed issue with DX12 DepthStencil having both DepthStencil AND SRV usage.

### DIFF
--- a/src/Vex/Resource.h
+++ b/src/Vex/Resource.h
@@ -2,6 +2,7 @@
 
 #include <span>
 
+#include <Vex/Concepts.h>
 #include <Vex/Handle.h>
 #include <Vex/Types.h>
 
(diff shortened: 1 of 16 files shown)


src/Vex/Resource.h
```diff
@@ -69,10 +71,11 @@ class ResourceMappedMemory
 };
 
 template <class T>
+    requires(not IsContainer<T>)
 void ResourceMappedMemory::SetData(const T& inData)
 {
     const u8* dataPtr = reinterpret_cast<const u8*>(&inData);
-    SetData(std::span{ dataPtr, sizeof(T) });
+    SetData(std::as_bytes(std::span{ dataPtr, sizeof(T) }));
```
alexandreLBarrett, line 78: Maybe we could do one that works with containers?

alexandreLBarrett (approved):

PR #80, opened 2025-09-27
Added automatic unit test script

Narvin-Chana:
- Added GTest to the project for unit testing of Vex features.
- Added a new tests/ folder with appropriate CMake file to run GTest for a specific target.
- Added a python script that runs the tests for each available permutation.
- Added some sample tests, the first one tests the Fact function. The second will create a GfxBackend, and will only fail upon having a runtime crash.

This will have to be ran locally, since GitHub Actions do not have GPUs (I believe).
But it will allow for test code to no longer go through examples, but instead actual unit test code! This should keep examples simpler and more "instructive" and also allow for more twisted cases in tests.

The script is in the newly created 'scripts/' folder, but should be ran from the Vex root, by calling `./scripts/VexTests.py`


scripts/VexTests.py
```diff
@@ -0,0 +1,197 @@
...
+def check_msvc_requirements():
+    """Check if MSVC is available, warn if not"""
+    if platform.system() != "Windows":
+        return False
+
+    try:
+        subprocess.run(['cl.exe'], capture_output=True, check=False)
+        return True
+    except FileNotFoundError:
+        print("⚠️  MSVC compiler (cl.exe) not found in PATH")
```
alexandreLBarrett, line 28: This is for sure some Claude emoji

src/Vulkan/VkDebug.h
```diff
@@ -11,11 +11,19 @@ VKAPI_ATTR VkBool32 VKAPI_CALL debugCallback(::vk::DebugUtilsMessageSeverityFlag
...
 #if !VEX_SHIPPING
 // Only pass Vk types
 template <class T>
 void SetDebugName(::vk::Device device, const T& object, const char* name)
 {
+    if (!GEnableDebugName)
+    {
+        return;
+    }
+
```
alexandreLBarrett, line 26: I'm not sure if I prefer this to be here or at call sites. Feels sneaky to have a function that doesn't do what you want because of some global variable. I agree though that we shouldn't set names when not in debug mode.

scripts/VexTests.py
Narvin-Chana: yep, msvc is forked

src/Vulkan/VkDebug.h
alexandreLBarrett: Fair enough

PR #155, opened 2026-03-24
Add Compile context to store Virtual include files and files to compile

MartinDew:
Add a new compile context object that stores virtual files. 
Optional loadModule function that may need to be removed. 

TODO : 
Remove source code from ShaderKey and make sure slang is able to compile from VCS


examples/hello_slang_modules/HelloSlangModule.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+        // Enqueue CPU readback — internally adds a barrier + copy to a
+        // staging buffer; the actual data becomes available after Submit.
+        vex::BufferReadbackContext readback = ctx.EnqueueDataReadback(outputBuffer);
+
+        vex::SyncToken token = graphics.Submit(ctx);
+        graphics.WaitForTokenOnCPU(token);
+
+        // Retrieve the data from the staging buffer into our CPU vector.
+        readback.ReadData(
```
alexandreLBarrett: nit: you could use std::as_writeable_bytes on your vector. It will save you all these casts and explicit creations

src/Vex/Shaders/DXCImpl.cpp
```diff
@@ -173,6 +185,70 @@ std::vector<std::pair<std::wstring, std::wstring>> BuildDefineList(const ShaderK
...
+    ULONG STDMETHODCALLTYPE AddRef() override
+    {
+        return ++refCount;
+    }
+    ULONG STDMETHODCALLTYPE Release() override
+    {
+        ULONG count = --refCount;
+        if (count == 0)
+        {
+            delete this;
```
alexandreLBarrett: Although valid c++. This feels like a code smell. Have you seen this pattern used in examples?

src/Vex/Shaders/ShaderCompiler.cpp
```diff
@@ -39,23 +39,35 @@ ShaderCompiler::ShaderCompiler(const ShaderCompilerSettings& compilerSettings)
...
+
 ShaderEnvironment ShaderCompiler::CreateShaderEnvironment()
 {
     ShaderEnvironment env;
     env.defines.emplace_back("VEX_DEBUG", std::to_string(VEX_DEBUG));
     env.defines.emplace_back("VEX_DEVELOPMENT", std::to_string(VEX_DEVELOPMENT));
     env.defines.emplace_back("VEX_SHIPPING", std::to_string(VEX_SHIPPING));
-    env.defines.emplace_back("VEX_RAYTRACING", std::to_string(GPhysicalDevice->IsFeatureSupported(Feature::RayTracing)));
+    env.defines.emplace_back("VEX_RAYTRACING",
+                             std::to_string(GPhysicalDevice->IsFeatureSupported(Feature::RayTracing)));
```
alexandreLBarrett: unnecessary change

alexandreLBarrett (commented):
The proposed feature is justified but the implementation needs work. Many decisions regarding the internal implementation were done without talking to us and checking if the implementation is according to our global design.

alexandreLBarrett (commented):
As a general design question. What is the purpose of the ShaderCompilerContext? Its existence is tied to the ShaderCompiler so why have an external class for that? why not just store the "VFS" in the compiler itself and then pass it to the impls when needed. It would remove the opaqueImpl concept which is confusing and remove unnecessary abstractions

PR #156, opened 2026-03-27
Vulkan Ray Tracing

alexandreLBarrett:
This PR implements hardware ray tracing in the Vulkan RHI. It should now have an equivalent feature set as DX12's implementation. Most of the changes are in VkAccelerationStructure, VkRayTracingPipelineState and CommandContext.

All of the RT tests pass without issues in both RHIs.

Some code is very similar from DX to VK but I think its not worth the hassle to combine them. Maybe in the future if the need rises. Let me know if you disagree and I'll combine some things.

### DIFF
--- a/src/Vex/CommandContext.cpp
+++ b/src/Vex/CommandContext.cpp
@@ -1078,7 +1078,7 @@ void CommandContext::BuildTLAS(const AccelerationStructure& accelerationStructur
     Buffer instanceBuffer = CreateTemporaryBuffer({
         .name = accelStruct.GetDesc().name + "_build_tlas_instances",
         .byteSize = instanceData.size(),
-        .usage = BufferUsage::ShaderRead | BufferUsage::BuildAccelerationStructure,
+        .usage = BufferUsage::BuildAccelerationStructure,
     });
     RHIBuffer& rhiInstanceBuffer = graphics->GetRHIBuffer(instanceBuffer.handle);
     EnqueueDataUpload(instanceBuffer, instanceData);
--- a/src/Vex/CommandContext.cpp
+++ b/src/Vex/CommandContext.cpp
@@ -908,6 +908,21 @@ void CommandContext::BuildBLAS(const AccelerationStructure& accelerationStructur
         u32 transformIndex = 0;
         for (const BLASGeometryDesc& blasGeometry : desc.geometry)
         {
+            // TODO(https://trello.com/c/srGndUSP): Handle other vertex formats, this should be cross-referenced
+            // with Vulkan to make sure only formats supported by both APIs are accepted.
+            if (*blasGeometry.vertexBufferBinding.strideByteSize > sizeof(float) * 3)
+            {
+                VEX_LOG(
+                    Warning,
+                    "Vex currently does not support acceleration structure geometry whose vertices have a format "
+                    "different to 12 bytes (RGB32). Your vertex buffer binding has a different stride than this, this "
+                    "is ok as long as the user is aware that elements outside the first 12 bytes will be ignored.");
+            }
+
+            VEX_ASSERT(*blasGeometry.vertexBufferBinding.strideByteSize >= sizeof(float) * 3,
+                       "Vex currently does not support acceleration structure geometry whose vertices have a stride "
+                       "smaller than 12 bytes.");
+
             RHIBLASGeometryDesc rhiBLASGeometry{
                 .vertexBufferBinding =
                     RHIBufferBinding(blasGeometry.vertexBufferBinding,
@@ -917,6 +932,8 @@ void CommandContext::BuildBLAS(const AccelerationStructure& accelerationStructur
 
             if (blasGeometry.indexBufferBinding.has_value())
             {
+                VEX_CHECK(blasGeometry.indexBufferBinding->strideByteSize == sizeof(u32),
+                          "Vex only supports 32bit index types")
                 rhiBLASGeometry.indexBufferBinding =
                     RHIBufferBinding(*blasGeometry.indexBufferBinding,
                                      graphics->GetRHIBuffer(blasGeometry.indexBufferBinding->buffer.handle));
@@ -924,7 +941,7 @@ void CommandContext::BuildBLAS(const AccelerationStructure& accelerationStructur
 
             if (blasGeometry.transform.has_value())
             {
-                rhiBLASGeometry.transform = RHIBufferBinding{
+                rhiBLASGeometry.transformBufferBinding = RHIBufferBinding{
                     .binding = { .buffer = transformBuffer,
                                  .offsetByteSize = TransformMatrixSize * transformIndex,
                                  .rangeByteSize = TransformMatrixSize },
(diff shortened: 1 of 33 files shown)


Narvin-Chana (changes requested):
Good stuff!

src/Vex/CommandContext.cpp
```diff
@@ -913,10 +931,10 @@ void CommandContext::BuildBLAS(const AccelerationStructure& accelerationStructur
         .name =
             graphics->GetRHIAccelerationStructure(accelerationStructure.handle).GetDesc().name + "_build_blas_scratch",
         .byteSize = buildInfo.scratchByteSize,
-        .usage = BufferUsage::ScratchBuffer | BufferUsage::ReadWriteBuffer,
+        .usage = BufferUsage::AccelerationStructure | BufferUsage::ScratchBuffer | BufferUsage::ReadWriteBuffer,
     };
     Buffer scratchBuffer = CreateTemporaryBuffer(scratchBufferDesc);
-    Barrier(scratchBuffer, RHIBarrierSync::BuildAccelerationStructure, RHIBarrierAccess::ShaderReadWrite);
+    Barrier(scratchBuffer, RHIBarrierSync::BuildAccelerationStructure, RHIBarrierAccess::AccelerationStructureWrite);
```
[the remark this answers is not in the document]
alexandreLBarrett: I think it was something related to vulkans needs. Something like any buffer used for AS setup must have the AccelerationStructure usage. I'll double check

src/Vulkan/RHI/VkPipelineState.cpp
```diff
@@ -255,33 +258,192 @@ void VkComputePipelineState::Cleanup(ResourceCleanup& resourceCleanup)
...
+
+        return VEX_VK_CHECK <<= ctx->device.createShaderModuleUnique(shaderModulecreateInfo);
+    };
+
+    std::vector<::vk::UniqueShaderModule> modules;
+    std::vector<::vk::PipelineShaderStageCreateInfo> stages;
+    std::vector<::vk::RayTracingShaderGroupCreateInfoKHR> groups;
+
+    using VkShaderGroupCreateInfo = ::vk::RayTracingShaderGroupCreateInfoKHR;
+    auto registerShaderStage = [&](const std::vector<NonNullPtr<Shader>>& shaders,
```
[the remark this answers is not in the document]
alexandreLBarrett: even for ones on the stack?

tests/shaders/RayTracingAABB.slang
```diff
@@ -1,5 +1,13 @@
 import Vex;
 
+struct UniformStruct
+{
+    uint outputHandle;
+    uint accelerationStructureHandle;
+};
+
+[vk::push_constant] uniform UniformStruct Uniforms;
```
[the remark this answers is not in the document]
alexandreLBarrett: Not to my knowledge

tests/AccelerationStructureTest.cpp
```diff
@@ -622,7 +614,7 @@ TEST_P(ASAABBTest, CreateAABBTraceShader)
         },
         ConstantBinding(data),
         {
-            1, 1, 1
+            100, 100, 1
```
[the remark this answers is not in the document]
alexandreLBarrett: I think i wanted to write the result to a texture for better visualization

src/Vulkan/RHI/VkPipelineState.cpp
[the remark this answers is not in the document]
Narvin-Chana: Yeah, up to now all lambdas (even local scope ones) were PascalCase. We could change the convention but might as well stick with it since changing would require going over all lambdas.

src/Vulkan/RHI/VkPipelineState.cpp
```diff
@@ -255,32 +259,216 @@ std::unique_ptr<RHIComputePipelineState> VkComputePipelineState::Cleanup()
...
 {
-    VEX_NOT_YET_IMPLEMENTED();
 }
 
 std::vector<MaybeUninitialized<RHIBuffer>> VkRayTracingPipelineState::Compile(
     const RayTracingShaderCollection& shaderCollection, RHIResourceLayout& resourceLayout, RHIAllocator& allocator)
 {
-    VEX_NOT_YET_IMPLEMENTED();
-    return {};
+    auto createShaderModule = [&](const Shader& s)
```
Narvin-Chana: createShaderModule -> CreateShaderModule

Narvin-Chana (commented):
LGTM, there's just the issue with crashing on Vulkan HWRT when making multple tlas/blas that remains.

src/Vex/CommandContext.cpp
```diff
@@ -1035,29 +1067,42 @@ void CommandContext::BuildTLAS(const AccelerationStructure& accelerationStructur
...
-    BufferDesc scratchBufferDesc{
-        .name =
-            graphics->GetRHIAccelerationStructure(accelerationStructure.handle).GetDesc().name + "_build_tlas_scratch",
+    RHIAccelerationStructure& accelStruct = graphics->GetRHIAccelerationStructure(accelerationStructure.handle);
+    std::vector<std::byte> instanceData = accelStruct.GetInstanceBufferData(rhiTLASDesc);
+
+    Buffer instanceBuffer = CreateTemporaryBuffer({
+        .name = accelStruct.GetDesc().name + "_build_tlas_instances",
+        .byteSize = instanceData.size(),
+        .usage = BufferUsage::ShaderRead | BufferUsage::BuildAccelerationStructure,
```
Narvin-Chana: Could you test without the ShaderRead flag? It might not even be necessary.

src/Vulkan/RHI/VkAccelerationStructure.cpp
```diff
@@ -1,28 +1,284 @@
...
+
+    for (const RHIBLASGeometryDesc& geom : desc.geometries)
+    {
+        ::vk::AccelerationStructureGeometryKHR geometry{};
+        if (desc.type == ASGeometryType::Triangles)
+        {
+            VEX_ASSERT(geom.vertexBufferBinding);
+            VEX_ASSERT(geom.vertexBufferBinding->binding.strideByteSize);
+
+            static constexpr u32 vertexByteStride = sizeof(float) * 3;
```
Narvin-Chana: vertexByteStride -> VertexByteStride

src/Vulkan/RHI/VkPipelineState.cpp
```diff
@@ -255,32 +259,216 @@ std::unique_ptr<RHIComputePipelineState> VkComputePipelineState::Cleanup()
...
         return nullptr;
     }
-    VEX_NOT_YET_IMPLEMENTED();
+    auto cleanupPSO = std::make_unique<VkRayTracingPipelineState>(key, ctx, psoCache);
+    std::swap(cleanupPSO->rtPipeline, rtPipeline);
+
+    std::swap(cleanupPSO->groupHitTable, groupHitTable);
+    std::swap(cleanupPSO->rayCallableTable, rayCallableTable);
+    std::swap(cleanupPSO->rayGenTable, rayGenTable);
+    std::swap(cleanupPSO->rayMissTable, rayMissTable);
```
Narvin-Chana: Is only using `if (!rtPipeline)` valid? Is it possible that a shader table could be defined but not the pipeline?
If not we could maybe add asserts to ensure this.

src/Vulkan/RHI/VkTexture.cpp
```diff
@@ -333,10 +333,21 @@ void VkTexture::CreateImage(RHIAllocator& allocator)
...
+    if (desc.type == TextureType::TextureCube)
+    {
+        flags |= ::vk::ImageCreateFlagBits::eCubeCompatible;
+    }
+
     const bool needsConcurrent = ctx->queueFamilyIndices.size() > 1;
     ::vk::ImageCreateInfo createInfo{
         .pNext = useMutableFormat ? &imageFormatList : nullptr,
-        .flags = useMutableFormat ? ::vk::ImageCreateFlagBits::eMutableFormat : ::vk::ImageCreateFlags{},
+        // TODO: Add cube eCubeCompatible to flags when type is cube
```
Narvin-Chana: Should this be removed? Also does this mean eCubeCompatible was never previously supported?

Narvin-Chana (commented):
Tested in my renderer and it seems like this mostly works. I'm still getting DXC errors when compiling to spirv due to the layout of one of my buffers, but tlas/blas build works. I'm assuming that shaders which use them will work as well. 

Just a few last code comments.

src/Vulkan/RHI/VkTexture.cpp
alexandreLBarrett: Never was and I didnt see the validation error. It worked regardless so i suppose driver didnt need it

src/Vulkan/RHI/VkPipelineState.cpp
alexandreLBarrett: technically yes as they are both created in the same function but I could be extra secure if you really think I should

Narvin-Chana (approved):
lgtm

PR #158, opened 2026-04-06
Shader Compiler and Bindless Samplers

Narvin-Chana:
- Moves the shader compiler outside of Vex.
- The shader compiler can now be disabled from cmake (eg: for shipping builds or if the user wants to implement their own).
- Made DXC also be optional (meaning the user could use the shader compiler as a simple cache for precompiled shaders, without dxc and slang).
- Added a tool to generate built-in shaders for MipGeneration since we no longer can compile shaders from inside Vex. It generates a header file with a constexpr ShaderView for both spirv and dxil targets, for each resource type.
- The shader compiler now takes in a compilation target, making it agnostic to VEX_DX12/VEX_VULKAN, and/or the GPhysicalDevice (reflection however still depends on these, we should look into reworking this in the future anyway).
- Shader hot-reload was improved with additional options.
- Since the built-in shader had to be fully pre-compiled, the static sampler binding fiasco reach a critical point (we used to add the built-in sampler after the user's samplers and pass its register via define). To fix this I added bindless samplers to the project, allowing for us to get and dereference bindless handles for SamplerStates in shaders.
- Bindless samplers **are limited to 2048 simultaneous variants** (DX12 restriction), and on the Vulkan side lifespan is not yet handled, meaning that they are currently not able to be destroyed/recycled. This should be a fine restriction for most sane applications.
- Static samplers are still useable, but are entirely in the user-space. Vex will never use them for internal passes (eg: mip gen).


src/DX12/DX12Sampler.cpp
```diff
@@ -87,19 +88,39 @@ std::vector<D3D12_STATIC_SAMPLER_DESC> GetDX12StaticSamplersFromTextureSamplers(
             .AddressW = GetDX12TextureAddressModeFromAddressMode(samplers[i].addressW),
             .MipLODBias = samplers[i].mipLODBias,
             .MaxAnisotropy = samplers[i].maxAnisotropy,
-            .ComparisonFunc = GraphicsPipeline::GetD3D12ComparisonFuncFromCompareOp(samplers[i].compareOp),
+            .ComparisonFunc = GetD3D12ComparisonFuncFromCompareOp(samplers[i].compareOp),
```
alexandreLBarrett: nit: D3D12 prefix here where others have DX12. I would use the same everywhere

src/DX12/DX12Sampler.cpp
```diff
@@ -87,19 +88,39 @@ std::vector<D3D12_STATIC_SAMPLER_DESC> GetDX12StaticSamplersFromTextureSamplers(
...
+        .AddressU = GetDX12TextureAddressModeFromAddressMode(sampler.addressU),
+        .AddressV = GetDX12TextureAddressModeFromAddressMode(sampler.addressV),
+        .AddressW = GetDX12TextureAddressModeFromAddressMode(sampler.addressW),
+        .MipLODBias = sampler.mipLODBias,
+        .MaxAnisotropy = sampler.maxAnisotropy,
+        .ComparisonFunc = GetD3D12ComparisonFuncFromCompareOp(sampler.compareOp),
+        .MinLOD = sampler.minLOD,
+        .MaxLOD = sampler.maxLOD,
+    };
+    std::memcpy(samplerDesc.BorderColor, sampler.borderColor.data(), sizeof(float) * 4);
```
alexandreLBarrett: nit: could use std::uninitialized_copy_n here

src/RHI/RHIPipelineState.h
```diff
@@ -1,87 +1,60 @@
...
 {
 public:
-    using Key = GraphicsPipelineStateKey;
+    using Key = GraphicsPSOKey;
 
-    RHIGraphicsPipelineStateBase(const Key& key)
-        : key{ key }
+    RHIGraphicsPipelineStateBase(Key  key)
+        : key{std::move( key )}
     {
```
alexandreLBarrett:
```suggestion
    RHIGraphicsPipelineStateBase(Key key)
        : key{std::move(key)}
{
```

src/RHI/RHIPipelineState.h
```diff
@@ -1,87 +1,60 @@
...
 class RHIComputePipelineStateBase
 {
 public:
-    using Key = ComputePipelineStateKey;
+    using Key = ComputePSOKey;
 
-    RHIComputePipelineStateBase(const Key& key)
-        : key{ key }
+    RHIComputePipelineStateBase(Key  key)
+        : key{std::move( key )}
```
alexandreLBarrett:
```suggestion
    RHIComputePipelineStateBase(Key key)
        : key{std::move(key)}
```

src/ShaderCompiler/Compiler/DXC/DXCCompiler.cpp
```diff
@@ -0,0 +1,407 @@
...
+    wStringPaths.reserve(includeDirectories.size() + 1);
+    for (auto& additionalIncludeFolder : includeDirectories)
+    {
+        wStringPaths.emplace_back(StringToWString(additionalIncludeFolder.string()));
+        args.insert(args.end(), { L"-I", wStringPaths.back().c_str() });
+    }
+
+    // Also adds the current working directory for the Vex.hlsli file.
+    wStringPaths.emplace_back(StringToWString(std::filesystem::current_path().string()));
+    args.insert(args.end(), { L"-I", wStringPaths.back().c_str() });
```
alexandreLBarrett, line 180: isnt this implicit with the default include handler?

src/ShaderCompiler/Compiler/Slang/SlangCompiler.cpp
```diff
@@ -20,19 +22,30 @@ namespace SlangImpl_Internal
...
                                                       diagnostics.writeRef())) ||
-        !bytecodeBlob || diagnostics)
+        !bytecodeBlob)
+    {
+        if (diagnostics)
+        {
+            return std::unexpected(std::format("Failed to get compiled shader bytecode: {}.",
+                                               static_cast<const char*>(diagnostics->getBufferPointer())));
+        }
+        return std::unexpected("Failed to compile shader.");
```
alexandreLBarrett: nit: could mention that there were no diagnostics available or something like that

src/ShaderCompiler/ShaderCompiler.cpp
```diff
@@ -0,0 +1,376 @@
...
+    Shader& shader = *GetShader(key, false);
+    CompilerBase& compiler = GetCompiler(key);
+    return HandleCompiledShader(shader, compiler.CompileShader(shader, sourceCode, globalShaderEnv, compilerSettings));
+}
+
+void ShaderCompiler::SetShaderCompilationErrorsCallback(ShaderHotReloadErrorsCallback&& callback)
+{
+    if (!compilerSettings.enableShaderHotReload)
+    {
+        VEX_LOG(Warning, "Cannot set shader compilation errors callback when not in shader hot-reload mode...");
```
alexandreLBarrett: nit: I would say "this has no effect" more than "Cannot set"

src/ShaderCompiler/ShaderCompiler.cpp
```diff
@@ -0,0 +1,376 @@
...
+    VEX_LOG(Info, "Recompiled all filepath-based shaders ({})...", numRecompiledShaders);
+}
+
+void ShaderCompiler::RecompileShaders(const Span<const ShaderKey> shaderKeys)
+{
+    const u32 numRecompiledShaders = HotReloadShaders(shaderKeys);
+    VEX_LOG(Info, "Recompiled the passed-in filepath-based shaders ({})...", numRecompiledShaders);
+}
+
+std::optional<std::filesystem::path> ShaderCompiler::ConvertVirtualFilepathToFilepath(const ShaderKey& key)
```
alexandreLBarrett: How does this convert the path from a virtual one to a real one? It seems to me to be more like validation than conversion

src/ShaderCompiler/ShaderCompiler.cpp
```diff
@@ -0,0 +1,376 @@
...
+                "Invalid shader compiler backend when attempting to obtain compiler, extension: {} with key: {}.",
+                extension,
+                key.filepath);
+        std::unreachable();
+    }
+}
+
+NonNullPtr<Shader> ShaderCompiler::GetShader(const ShaderKey& key, bool allowHotReload)
+{
+    if (const auto el = shaderCache.find(key); el != shaderCache.end())
```
alexandreLBarrett: nit: use more significant name. Maybe shader?

src/ShaderCompiler/ShaderCompiler.cpp
```diff
@@ -0,0 +1,376 @@
...
+    if (compilerSettings.dumpShaderOutputBytecode)
+    {
+        const std::vector<byte>& shaderBytecode = compilationResult->compiledCode;
+        std::string uniqueFilename = std::format("{}_{}_{}_{}_{}",
+                                                 shader.GetHash(),
+                                                 shader.GetKey().entryPoint,
+                                                 shader.GetKey().defines,
+                                                 shader.GetKey().type,
+                                                 shader.GetKey().compiler);
+        // Ensure the filepath isn't too long (OS limit is usually ~250, but we also have to store the current_path()).
```
alexandreLBarrett, line 257: We could store two files. One for the shader compile env and one for the actual byte code. and differentiate similar shaders with a hash. It would prevent the path length issue and be more readable for us. Its not critical at all. Just for future improvement

src/ShaderCompiler/ShaderCompiler.h
```diff
@@ -0,0 +1,130 @@
...
+};
+
+using ShaderHotReloadErrorsCallbackFuncType =
+    ShaderHotReloadErrorResponse(Span<const std::pair<ShaderKey, std::string>>);
+using ShaderHotReloadErrorsCallback =
+#ifdef __cpp_lib_move_only_function
+std::move_only_function<ShaderHotReloadErrorsCallbackFuncType>;
+#else
+std23::move_only_function<ShaderHotReloadErrorsCallbackFuncType>;
+#endif
```
alexandreLBarrett: couldn't we abstract this in some way? Like hide it in a header. That way it will be trivial to remove when compilers will be up to date and it will help with code readability

src/ShaderCompiler/ShaderCompiler.h
```diff
@@ -0,0 +1,130 @@
...
+    void SetShaderCompilationErrorsCallback(ShaderHotReloadErrorsCallback&& callback);
+
+    // Recompiles all filepath-based shaders which have changed since the last compilation. Useful for shader
+    // development and hot-reloading. You generally want to avoid calling this too often if your application has many
+    // shaders. Only valid if enableShaderHotReload is enabled.
+    void RecompileChangedShaders();
+    // Recompiles all filepath-based shaders, could cause a big hitch depending on how many shaders your application
+    // uses. Shaders compiled from source Only valid if enableShaderHotReload is enabled.
+    void RecompileAllShaders();
+    // Recompiles all filepath-based shaders passed-in.
```
alexandreLBarrett, line 92:
```suggestion

    // Recompiles all filepath-based shaders, could cause a big hitch depending on how many shaders your application
    // uses. Only valid if enableShaderHotReload is enabled.
    void RecompileAllShaders();
    
    // Recompiles all filepath-based shaders passed-in.
```

src/ShaderCompiler/ShaderCompilerSettings.h
```diff
@@ -0,0 +1,54 @@
...
+    spirv_1_6
+};
+
+struct ShaderCompilerSettings
+{
+    // Determines whether to compile to the dxil or spirv target.
+    CompilationTarget target = VEX_DX12 ? CompilationTarget::DXIL : CompilationTarget::SPIRV;
+
+    // Determines the shader model to use, uses the minimum required Vex shader model by default.
+    ShaderModel shaderModel = ShaderModel::SM_6_7;
```
alexandreLBarrett, line 34: Why do we allow SM to be lower than 6_7? It would make the shader incompatible with Vex. IMO we should remove anything that isnt supported to remove confusion

src/ShaderCompiler/ShaderCompilerSettings.h
```diff
@@ -0,0 +1,54 @@
...
+
+    // When compiling to spirv, defines the spirv version to use, uses the minimum required Vex spirv version by default.
+    SpirvVersion spirvVersion = SpirvVersion::spirv_1_6;
```
alexandreLBarrett, line 37: Same here perhaps? I'm not sure what's the minimum spirv version we require

src/Vex/Buffer.h
```diff
@@ -29,6 +29,7 @@ BEGIN_VEX_ENUM_FLAGS(BufferUsage, u16)
     IndexBuffer                 = 1 << 4,
     // Buffers used as parameters for an indirect dispatch.
     IndirectArgs                = 1 << 5,
+    // TODO: check/rework this, error prone
```
alexandreLBarrett: fix todo or link trello

src/Vex/Graphics.cpp
```diff
@@ -512,6 +413,30 @@ std::vector<BindlessHandle> Graphics::GetBindlessHandles(Span<const ResourceBind
...
 }
 
+BindlessHandle Graphics::GetBindlessSampler(const BindlessTextureSampler& sampler)
+{
+    if (bindlessSamplers.size() == GMaxBindlessSamplerCount)
+    {
+        VEX_LOG(
+            Fatal,
+            "Max number of different bindless samplers reached (2048). You must reduce the number of variations used.");
+    }
```
alexandreLBarrett, line 429: Shouldn't this be validated when creating a sampler and not when getting one?

src/Vex/PipelineStateCache.h
```diff
@@ -17,38 +15,31 @@ class Graphics;
...
     PipelineStateCache(PipelineStateCache&&) = default;
     PipelineStateCache& operator=(PipelineStateCache&&) = default;
 
-    RHIResourceLayout& GetResourceLayout();
+    RHIGraphicsPipelineState* GetGraphicsPipelineState(const DrawDesc& drawDesc,
+                                                       const RenderTargetState& renderTargetState,
+                                                       std::unique_ptr<RHIGraphicsPipelineState>& oldPSO);
+    RHIComputePipelineState* GetComputePipelineState(const ShaderView& computeShader,
+                                                     std::unique_ptr<RHIComputePipelineState>& oldPSO);
+    RHIRayTracingPipelineState* GetRayTracingPipelineState(const RayTracingShaderCollection& shaderCollection,
```
alexandreLBarrett, line 29: nit: design comment: we could return the old (along with the new) here instead of using out param? It's not major but it might improve the code clarity

src/Vex/ResourceReadbackContext.h
```diff
@@ -48,9 +48,9 @@ class TextureReadbackContext
     }
 
 private:
-    TextureReadbackContext(const Buffer& buffer,
+    TextureReadbackContext(Buffer buffer,
                            Span<const TextureRegion> textureRegions,
-                           const TextureDesc& textureDesc,
+                           TextureDesc textureDesc,
                            Graphics& graphics);
```
alexandreLBarrett, line 54: I suppose this is to allow for move into instead of copying? if so did you make sure we move into it whenever we can?

src/Vulkan/VkSampler.cpp
```diff
@@ -0,0 +1,44 @@
...
+{
+    ::vk::SamplerCreateInfo vkSamplerCreateInfo = FillInVkSamplerCreateInfoWithTextureSamplerBase(sampler);
+    vkSamplerCreateInfo.borderColor = BorderColorToVkBorderColor(sampler.borderColor);
+    return vkSamplerCreateInfo;
+}
+::vk::SamplerCreateInfo GraphicsPipeline::GetVkSamplerCreateInfoFromBindlessTextureSampler(
+    const BindlessTextureSampler& sampler)
+{
+    ::vk::SamplerCreateInfo vkSamplerCreateInfo = FillInVkSamplerCreateInfoWithTextureSamplerBase(sampler);
+    // TODO: use VK_EXT_custom_border_color
```
alexandreLBarrett: do it or trello it

CMakeLists.txt
```diff
@@ -29,16 +31,16 @@ include(VexHelpers)
 # Will correctly install our custom githook to run clang-format on staged files when committing
 include(GitHooks)
 
-# Contains all logic to fetch and include DXC
-include(VexDXC)
-
+option(VEX_ENABLE_SHADER_COMPILER "Enables the possibility to use Vex's built-in ShaderCompiler." ON)
+option(VEX_ENABLE_DXC "Enables the possibility to use the DXC shader compiler backend." ON)
```
alexandreLBarrett, line 40: Why is DXC handled differently to slang here?

alexandreLBarrett (commented):
Overall very impressive refactor! This was a beast to review but its a good breath of fresh air for all of the shader compilation code. Good job!

CMakeLists.txt
Narvin-Chana: In what way? DXC is the current default compiler backend when no configuration is provided.
Otherwise both should be handled very similarly.

src/ShaderCompiler/Compiler/DXC/DXCCompiler.cpp
Narvin-Chana: Unfortunately not, we had this before as well.

Narvin-Chana: It only adds the path of the currently compiled file, not the .exe current path.

src/ShaderCompiler/ShaderCompilerSettings.h
Narvin-Chana: Well the default is something sane, there's not really a use for a higher shader model unless the user wishes to use more recent cutting-edge features.
For why we don't force it to something lower, its due to the shader compiler technically not being dependent on Vex (although in the end we still use a lot of Vex utility files and such...).
How should we force it to this? An assert inside the shader compiler?

src/Vex/PipelineStateCache.h
Narvin-Chana: I'll allow myself to ignore this one for now since its fully internal Vex logic, I agree that optional out params are a bit of a C code smell though.

src/ShaderCompiler/ShaderCompilerSettings.h
Narvin-Chana: The minimum we require is the version the shader compiler uses as a default value.

alexandreLBarrett (approved):
all good

PR #161, opened 2026-05-02
Experimental Vex C++20 module support + Enum flags rework

Narvin-Chana:
Vex now exposes a Vex.cppm module which users can use to consume Vex as a C++20 module:
- This module is by default NOT built unless the user sets VEX_USE_MODULES to true (and the underlying compiler + CMake combo also supports modules).
- The Vex.cppm module was hand-written, there might be a bunch of issues with certain symbols which are missing, these will be added when found.
- When building with the VEX_USE_MODULES cmake option, Vex will now use (and provide) the magic_enum module.
- While looking into magic_enum, I found that magic_enum comes with a built-in formatter specialization for compatible enum types, meaning I was able to remove a bit of our code relating to this and use their formatter directly instead.
- I had to bump Vex to use the master branch version of magic_enum which contains many module-related fixes that the latest release (from 2024), doesn't contain.
- Certain hacks/compiler specific tricks had to be used to get everything to play nicely with both approaches (headers-only and modules).
- Moved most of our macros into a VexMacros.h file, since modules cannot export macros.
- All examples/tools/tests were updated to also consume the VEX_USE_MODULES option, meaning that we can build these as consumers of headers and/or consumers of modules. Currently tests will only run one of the two, but there should hopefully be very little divergence between the two going forward (most issues are already fixed).
- Updated Slang and DXC to be cloned into versioned folders, making updating the version used auto-update (whereas before it might not auto-update).
- Updated the .README.

## Reasons behind certain code changes
In the .cppm module, we are required to re-export any symbols users would need when using Vex. This brought up a few issues due to namespaces NOT being exportable (you'd have to manually export each symbol inside the namespace or create another namespace to alias the first one):
- Updated vex to use a new form of enum flags (similar to vulkan's Flags<T> type), since the previous version used namespaces.
- Changed all our public collections of functions to be structs with static methods, instead of namespaces. This should be the new convention moving forwards as it makes C++20 module export way easier.
- Created struct collections for most public functions we have, to also avoid having to manually export methods one by one.

## The module support is marked as experimental due to multiple reasons:
1. CMake and Compiler support for C++20 modules can be flaky depending on the combination used. As an example CMake and Clang-CL don't play well together, even though Clang itself supports C++20 modules perfectly fine. These issues should hopefully be ironed out over time...
2. As mentioned above the Vex.cppm module could still be missing certain symbols/types.
3. This remains still untested on certain compilers (eg: GCC), hopefully our CI will be able to catch most issues here.

## ~~Remaining work to do:~~ DONE!
- [x] Create ByteUtil struct for byte functions (and others for free function libraries in the Vex folder).
- [x] Test out the implementation in my own renderer.
- [x] Investigate why the DX12 agility sdk version define is no longer valid.

And as an FYI, the changes I made to the "Installation rules" section in CMake HAS to be written in this annoying duplicated way...


src/Vex/GraphicsPipeline.h
```diff
@@ -224,19 +223,17 @@ enum class BlendOp : u8
...
 // Flags for which channels are written to the render target.
-BEGIN_VEX_ENUM_FLAGS(ColorWriteMask, u8)
+enum class ColorWriteMask : u8
+{
     None,
     Red = 1,
     Green = 2,
     Blue = 4,
     Alpha = 8,
     All = 0b1111,
```
alexandreLBarrett: nit: We should rewrite thos as bitshifts like all other flag enums. and All could be all of the possible values ORed (or 0xFF). TextureAspect enum is a good example

src/Vex.cppm
```diff
@@ -0,0 +1,113 @@
...
+using vex::byte;
+using vex::u8;
+using vex::u16;
+using vex::u32;
+using vex::u64;
+using vex::i8 ;
+using vex::i16;
+using vex::i32;
+using vex::i64;
+// clang-format on
```
alexandreLBarrett, line 128: was using clang format off that necessary here lol

CMakeLists.txt
```diff
@@ -112,7 +123,7 @@ message(STATUS "Fetching magic_enum...")
 FetchContent_Declare(
     magic_enum
     GIT_REPOSITORY https://github.com/Neargye/magic_enum.git
-    GIT_TAG v0.9.7
+    GIT_BRANCH master
```
alexandreLBarrett: I suppose there isnt a tag that suits our needs? We could use a commit number to prevent being affected by upstream changes

src/Vex.cppm
Narvin-Chana: ye.. it sorts it alphabetically, not alphanumerically

src/DX12/RHI/DX12PhysicalDevice.cpp
```diff
@@ -10,10 +10,10 @@
...
 {
 
-static constexpr D3D_SHADER_MODEL GMinimumShaderModel = D3D_SHADER_MODEL_6_6;
-static constexpr D3D12_RAYTRACING_TIER GMinimumRayTracingTier = D3D12_RAYTRACING_TIER_1_0;
-static constexpr D3D12_MESH_SHADER_TIER GMinimumMeshShaderTier = D3D12_MESH_SHADER_TIER_1;
-static constexpr D3D_FEATURE_LEVEL GMinimumFeatureLevel = D3D_FEATURE_LEVEL_12_1;
+inline constexpr D3D_SHADER_MODEL GMinimumShaderModel = D3D_SHADER_MODEL_6_6;
+inline constexpr D3D12_RAYTRACING_TIER GMinimumRayTracingTier = D3D12_RAYTRACING_TIER_1_0;
+inline constexpr D3D12_MESH_SHADER_TIER GMinimumMeshShaderTier = D3D12_MESH_SHADER_TIER_1;
+inline constexpr D3D_FEATURE_LEVEL GMinimumFeatureLevel = D3D_FEATURE_LEVEL_12_1;
```
alexandreLBarrett: Why inline here? Are these symbols used elsewhere from this file?

src/DX12/DX12Formats.h
```diff
@@ -9,7 +9,7 @@ namespace vex::dx12
 {
 
 // Convert from TextureFormat to DXGI_FORMAT
-constexpr inline DXGI_FORMAT TextureFormatToDXGI(TextureFormat format, bool isSRGB)
+constexpr DXGI_FORMAT TextureFormatToDXGI(TextureFormat format, bool isSRGB)
```
alexandreLBarrett, line 12: Wont this cause ODR if two cpps include that file?

src/ShaderCompiler/Compiler/DXC/DXCCompiler.cpp
```diff
@@ -23,7 +22,7 @@ namespace DXCCompiler_Internal
 static std::wstring GetTargetFromShaderType(const ShaderType type, const ShaderCompilerSettings& compilerSettings)
 {
     std::wstring highestSupportedShaderModel =
-        StringToWString(std::string(magic_enum::enum_name(compilerSettings.shaderModel)));
+        PlatformUtil::PlatformUtil::StringToWString(std::string(magic_enum::enum_name(compilerSettings.shaderModel)));
```
alexandreLBarrett: PlatformUtil::PlatformUtil ?

src/Vex/AccelerationStructure.h
```diff
@@ -25,24 +25,26 @@ enum class ASType : u8
...
 
 // Flags for Acceleration Structure building. Currently most of these are not yet implemented.
-BEGIN_VEX_ENUM_FLAGS(ASBuild, u8)
-	None			= 0,
-	AllowUpdate		= 1 << 0,	// Allows for incremental updates to the AccelerationStructure.
-	AllowCompaction = 1 << 1,	// Allows for acceleration structure compaction to save memory.
-	PreferFastTrace = 1 << 2,	// Optimizes building for raytracing performance. Incompatible with PreferFastBuild.
-	PreferFastBuild = 1 << 3,	// Optimizes building for build-speed. Incompatible with PreferFastTrace.
-	MinimizeMemory	= 1 << 4,	// Minimizes memory usage.
+enum class ASBuild : u8
```
alexandreLBarrett, line 28: Kind of a nit pick with the new enum flow. We could move all of the bit offset logic to the Flags class. The enum itself would have values incrementing from 0 to N and the Flags type would keep the bit version with BitN = 1 << EnumValueN. There would be a cleaner sparation between the flags and the enum values. idk food for thought

src/Vex/Logger.cpp
```diff
@@ -84,6 +87,27 @@ void Logger::CloseLogFile()
     logOutput.reset();
 }
 
+void Logger::LogToOutput(const std::string& logMessage)
+{
+    if (destinationFlags & LogDestination::Console)
+    {
+        std::println("{}", logMessage);
```
alexandreLBarrett: why not use std::cout << logMessage here instead. The format isnt used anyways

src/Vex.h
```diff
@@ -12,9 +12,10 @@
 #include <Vex/RayTracing.h>
 #include <Vex/TextureSampler.h>
 #include <Vex/Utility/ByteUtils.h>
-#include <Vex/Utility/Formattable.h>
 #include <Vex/Utility/NonNullPtr.h>
-#include <Vex/Utility/Validation.h>
+#if !VEX_MODULES
+#include <VexMacros.h>
+#endif
```
alexandreLBarrett: Not really necessary. It wont affect module code anyways no?

alexandreLBarrett (commented):
Nice job with the modules! Comments are mostly nitpick with one suggestion

src/DX12/DX12Formats.h
Narvin-Chana: constexpr functions and constexpr constructors are implicitly inline

src/Vex/AccelerationStructure.h
Narvin-Chana: The issue is that the ::None wouldn't exist, and we'd no longer be able to make enum values which use other flag values. Maybe this is a tradeoff we'd be ok with making? Since the ::None could just be Flags<T>{} (or FLags::None, if we really want). and for the using other flags, we could just use constexpr globals.
What do you think is best here?

src/Vex.h
Narvin-Chana: It's mainly for non-module users! So that they have everything included by just including Vex.h.

Narvin-Chana: Unless you meant the #if def?

alexandreLBarrett: Yeah i was talking about the #if def. its not needed since module dont export macros anyways

PR #164, opened 2026-06-19
Add an option to embed shaders instead of copying them.

MartinDew:
This creates some header files containing the content of the shaders and adds them to the build.


cmake/VexEmbedShaders.cmake
```diff
@@ -0,0 +1,44 @@
...
+
+    file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/src/Vex/Generated")
+    file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/src/Vex/Generated/EmbeddedShaders.h" "${_vex_embedded_header}")
+
+    target_include_directories(Vex PUBLIC "${CMAKE_CURRENT_BINARY_DIR}/src")
+
+    target_compile_definitions(Vex PRIVATE VEX_EMBEDDED_SHADERS=1)
+else()
+    target_compile_definitions(Vex PRIVATE VEX_EMBEDDED_SHADERS=0)
+endif()
```
Narvin-Chana: I like the idea, just a FYI MipGeneration.hlsl should not be touched.
It is used internally (aka by us) to generate pre-compiled bytecode for generating mips at runtime.

So if you want to further this PR, please make MipGeneration.hlsl not be changed.

Next how are the embeded shaders accessed once in the shader compiler? Is it an include at the start of the file? If so, just make sure to add a #line 0 afterwards, so user shader error messages aren't misaligned.

MartinDew: You mean Vex code itself doesn't use MipGeneration.hlsl?

alexandreLBarrett: Mip generation is already embedded in Vex code. We have a tool that bakes it as a cpp file.

see `src/Vex/Generated/MipGeneration.h` the hlsl is there only for reference and further updates but not for direct use

alexandreLBarrett: We could use this tho for mip generation and vex.hlsli for uniform use. wdyt @Narvin-Chana?

cmake/VexEmbedShaders.cmake
```diff
@@ -0,0 +1,40 @@
...
+        string(APPEND _vex_embedded_header
+            "inline constexpr const char* VexSlang = R\"VEX_SHADER_DELIM(\n${_vex_slang_src}\n)VEX_SHADER_DELIM\";\n\n")
+    endif()
+
+    string(APPEND _vex_embedded_header "} // namespace vex::embedded\n")
+
+    file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/src/Vex/Generated")
+    file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/src/Vex/Generated/EmbeddedShaders.h" "${_vex_embedded_header}")
+
+    target_include_directories(Vex PUBLIC "${CMAKE_CURRENT_BINARY_DIR}/src")
```
Narvin-Chana: Why is this here? I don't think the use of embedded shaders should add an entire include dir to the Vex target?

cmake/VexEmbedShaders.cmake
```diff
@@ -0,0 +1,40 @@
+option(VEX_USE_EMBEDDED_SHADERS "Whether Vex should embed helper shaders in its binary instead of distributing them in the bin folder." OFF)
+
+if (VEX_USE_EMBEDDED_SHADERS)
+
+    set(_vex_shader_deps "${VEX_ROOT_DIR}/shaders/MipGeneration.hlsl")
```
Narvin-Chana: I would remove MipGeneration.hlsl since its not a shader we ship to users.

cmake/VexEmbedShaders.cmake
```diff
@@ -0,0 +1,40 @@
...
+    if (VEX_ENABLE_SLANG)
+        file(READ "${VEX_ROOT_DIR}/shaders/Vex.slang" _vex_slang_src)
+        string(APPEND _vex_embedded_header
+            "inline constexpr const char* VexSlang = R\"VEX_SHADER_DELIM(\n${_vex_slang_src}\n)VEX_SHADER_DELIM\";\n\n")
+    endif()
+
+    string(APPEND _vex_embedded_header "} // namespace vex::embedded\n")
+
+    file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/src/Vex/Generated")
+    file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/src/Vex/Generated/EmbeddedShaders.h" "${_vex_embedded_header}")
```
Narvin-Chana: This code is only generated if the user's Vex uses the Shader Compiler. Thus I would recommend moving the generated file into a "src/ShaderCompiler/Generated" folder rather than in src/Vex.

cmake/VexEmbedShaders.cmake
```diff
@@ -0,0 +1,40 @@
...
+    if (VEX_ENABLE_DXC)
+        file(READ "${VEX_ROOT_DIR}/shaders/Vex.hlsli" _vex_hlsl_src)
+        string(APPEND _vex_embedded_header
+            "inline constexpr const char* VexHLSL = R\"VEX_SHADER_DELIM(\n${_vex_hlsl_src}\n)VEX_SHADER_DELIM\";\n\n")
+    endif()
+
+    if (VEX_ENABLE_SLANG)
+        file(READ "${VEX_ROOT_DIR}/shaders/Vex.slang" _vex_slang_src)
+        string(APPEND _vex_embedded_header
+            "inline constexpr const char* VexSlang = R\"VEX_SHADER_DELIM(\n${_vex_slang_src}\n)VEX_SHADER_DELIM\";\n\n")
```
Narvin-Chana: I would use std::string_view instead (requires also adding the include at the top of the generated file).

Narvin-Chana (changes requested):
Just some small changes to address, the global change looks good to me.

tests/ClearTests.cpp
```diff
@@ -1,5 +1,7 @@
 ﻿#include "VexTest.h"
 
+#include "Vex/Texture.h"
+
```
Narvin-Chana: Remove this change.

cmake/VexEmbedShaders.cmake
Narvin-Chana: You left MipGeneration.hlsl in _vex_shader_deps.

tests/ClearTests.cpp
MartinDew: You do know it doesn't compile without right?

cmake/VexEmbedShaders.cmake
MartinDew: This is because we generate the file in the build dir so we don't pollute the src dir. We have to add the include dir as public so that whoever consumes the project can also use that include dir and thus the generated file.

Narvin-Chana: Oooooh you generate the file inside the binary output, not inside the source dir! Makes sense! Then should the folder be called "src/...", maybe it should just directly be "Vex/Generated/...", to make it clearer?
I don't know what the convention should be for this?

cmake/VexEmbedShaders.cmake
```diff
@@ -0,0 +1,45 @@
...
+
+    file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/gen/ShaderCompiler/Generated")
+    file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/gen/ShaderCompiler/Generated/EmbeddedShaders.h" "${_vex_embedded_header}")
+
+    target_include_directories(Vex PUBLIC
+        "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/gen>"
+        $<INSTALL_INTERFACE:include>)
+
+    install(FILES "${CMAKE_CURRENT_BINARY_DIR}/gen/ShaderCompiler/Generated/EmbeddedShaders.h"
+        DESTINATION include/ShaderCompiler/Generated)
```
Narvin-Chana: Ok this is my bad, I didn't understand that this is in the binary output dir, not the Vex source tree!
That makes it make more sense, then in that case we can go back to this path:
"${CMAKE_CURRENT_BINARY_DIR}/Vex/Generated/EmbeddedShaders.h"

Narvin-Chana (approved):
LGTM! Just wait for Alex's OK before merging please.

alexandreLBarrett (approved):

